PR #139, opened 2018-04-05
Gravatar

nvisal1:
Gravatar implementation completed.


src/app/cube/user-profile/gravatar-image/gravatar-image.component.html
```diff
@@ -0,0 +1,3 @@
+<p>
```
gshaw1997: Remove this file if not being used.

src/app/cube/user-profile/gravatar-image/gravatar-image.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, OnInit, Input, Output } from '@angular/core';
+
+@Component({
+  selector: 'clark-gravatar-image',
+  template: `
+        <div id="container">
+            <img [src]="getGravatarImage()">
```
gshaw1997: getGravatarImage method is commented out in the component class, so it does not exist.

gshaw1997: So this method was moved to user profile component. Is gravatar-image.component.ts still in use? If not please remove.

src/app/cube/user-profile/user-profile.component.ts
```diff
@@ -35,6 +38,7 @@ export class UserProfileComponent implements OnInit {
     this.route.data.subscribe(val => {
```
gshaw1997: Be sure to unsubscribe from subscriptions when the component is destroyed.

gshaw1997: You can also avoid a subscription all together by grabbing the data from route snapshot if the data isn't expected to change during the lifetime of the component.

src/app/cube/user-profile/md5.ts
```diff
@@ -0,0 +1,231 @@
+// Taken from https://github.com/killmenot/webtoolkit.md5
```
gshaw1997: If a package is available, try to utilize it instead of copying code from an external js file into the project.

But try to avoid installing bulky packages if functionality needed is minimal and can be developed in house.

Also if there is a general function that may/can be used elsewhere in the project, moved to shared instead of housing it in component's directory.

A md5 hash package is available here: https://www.npmjs.com/package/md5

PR #226, opened 2018-04-26
Validation hotfix - Beta 16.13

gshaw1997:


src/app/core/auth.service.ts
```diff
@@ -34,6 +34,7 @@ export class AuthService {
           this.changeStatus(true);
         },
         error => {
+          this.cookies.remove('presence');
```
gshaw1997: This is the only real change. Other changes are just prettier formatting

PR #261, opened 2018-05-31
Add features for user bio and password change

nvisal1:


src/app/core/user.service.ts
```diff
@@ -22,6 +22,18 @@ export class UserService {
       .toPromise();
   }
 
+  changePassword(editPassword: object): Promise<any> {
```
gshaw1997: Remove method from production code if not in use.

PR #273, opened 2018-06-04
Display bio on user profile page

nvisal1:


src/app/cube/user-profile/user-information/user-information.component.scss
```diff
@@ -21,6 +21,11 @@
   }
 }
 
+.bio {
+  color: $dark-text;
```
nrwinner: Add a max-width property to prevent long single-line bios

PR #322, opened 2018-07-10
1.1.2 Hotfix - import fromEvent and debounceTime directly

nrwinner:


src/app/auth/register/personal-info/personal-info.component.ts
```diff
@@ -36,9 +40,13 @@ export class PersonalInfoComponent implements OnInit, OnDestroy {
   constructor(private auth: AuthService, private register: RegisterComponent) {}
 
   ngOnInit() {
```
DonSeannelly: If this method is not being used, it should be removed (along with the import and implement)

src/app/auth/register/personal-info/personal-info.component.ts
```diff
@@ -36,9 +40,13 @@ export class PersonalInfoComponent implements OnInit, OnDestroy {
...
+
+  }
+
+  ngAfterViewInit() {
     // listen for input events on the income input and send text to suggestion component after 650 ms of debounce
     this.subs.push(Observable.fromEvent(this.emailInput.nativeElement, 'input')
-      .map(x => x['currentTarget'].value)
+      .map(x =>  x['currentTarget'].value)
```
DonSeannelly: There's two spaces here. Try using the prettier plugin (or something similar) to take care of this for you.

We'll get it automated soon 🤞

src/app/shared/mappings-filter/mappings-filter.component.ts
```diff
@@ -9,17 +9,17 @@ import {
   HostListener,
   Input,
   OnChanges,
-  SimpleChanges,
-  IterableDiffers,
```
DonSeannelly: Thanks for removing unused imports 🙏

PR #353, opened 2018-08-20
Contributor name hotfix

nvisal1:
When selecting contributors in learning object builder, name should be displayed instead of username. 

Before merging, please pull and run builder.spec.js in e2e tests


src/app/onion/learning-object-builder/components/info-page/metadata/metadata.component.ts
```diff
@@ -96,20 +96,26 @@ export class LearningObjectMetadataComponent implements OnInit, OnDestroy {
   }
 
   addAuthor(index) {
+    console.log(this.arrayOfUsers[index]);
```
nrwinner: Please remove this dev log :heart:

PR #355, opened 2018-08-22
Updated Context Menu 

Ballen2713:
Updated the context menu to include a reset password option for the user


src/app/shared/navbar/navbar.component.ts
```diff
@@ -140,6 +143,10 @@ export class NavbarComponent implements OnInit, AfterContentChecked, OnDestroy {
   userprofile() {
     this.router.navigate(['users', this.authService.user.username]);
   }
+
```
DonSeannelly: Please add a comment to document this function

src/app/shared/navbar/navbar.component.ts
```diff
@@ -18,6 +18,9 @@ import * as md5 from 'md5';
 import { Subscription } from 'rxjs/Subscription';
 import { Subject } from 'rxjs/Subject';
 
+
```
DonSeannelly: Please remove the extra whitespace that was added

src/app/shared/navbar/navbar.component.ts
Ballen2713: On it!

package.json
```diff
@@ -27,11 +27,11 @@
     "@types/jest": "^23.1.1",
     "@types/uuid": "^3.4.3",
     "angular2-virtual-scroll": "^0.4.5",
-    "core-js": "^2.5.3",
-    "dropzone": "^5.4.0",
     "chai": "^4.1.2",
     "chance": "^1.0.16",
-    "cypress": "^3.0.2",
+    "core-js": "^2.5.3",
+    "cypress": "^3.1.0",
```
DonSeannelly: Didn't notice this until now, but it seems that we've had cypress listed as a regular dependency. Can you move it over to devDependencies so we can get that taken care of in the process of this PR?

src/app/shared/navbar/navbar.component.ts
```diff
@@ -137,12 +137,16 @@ export class NavbarComponent implements OnInit, AfterContentChecked, OnDestroy {
     this.authService.logout();
   }
 
-  //naviagtes to the user information by using the authservice class in the constructor above 
+  /**
+   * naviagtes to the users information by using the 
+   * @constructor authService
```
DonSeannelly: The `@constructor` tag is used to denote a function that intended to be called with the "new" keyword (see http://usejsdoc.org/tags-class.html). With that tag taken out though, the comment looks good!

PR #364, opened 2018-08-28
Modify important string storage and update text

nrwinner:
Many components include strings/copy that need to be updated quickly and by someone who may not be familiar with the code base. To that end, new .copy.ts files are added to many components so that text can be changed easily. Also includes small grammatical and informational changes.


src/app/cube/browse/browse.component.html
```diff
@@ -2,8 +2,8 @@
   <div class="browse-columns">
     <div class="sidebar-wrapper" [ngClass]="{'active': filtersDownMobile}">
       <div class="column-title">
-        Filters
-        <div class="filters-clear-all" (click)="clearAllFilters()">Clear all filters</div>
+        {{ copy.FILTERS }}
```
DonSeannelly: Is this descriptive enough, or should it be something along the lines of `FILTERS_TITLE`

Issue #321, opened 2018-07-09
Add Organization to Learning Object Detail View


gshaw1997:

DonSeannelly:
The organization should also link directly to the org. page

DonSeannelly:
Closed via #374

PR #374, opened 2018-08-29
Reimplement organization link to details page header 

nvisal1:
In the details page header, the author's name should route to the respective profile page. The author's organization should route to the respective organization page. This applies for both learning object author and contributors.


DonSeannelly:
Closes #321

PR #420, opened 2018-09-24
Implement cookie banner

nrwinner:


src/app/clark.module.ts
```diff
@@ -22,7 +22,8 @@ import { UnsupportedComponent } from './unsupported.component';
     ClarkRoutingModule,
     SharedModule,
     CoreModule.forRoot(),
-    BrowserAnimationsModule
+    BrowserAnimationsModule,
+    SharedModule
```
DonSeannelly: Why has the `SharedModule` been added twice?

src/app/shared/cookies/cookies.component.html
```diff
@@ -0,0 +1,13 @@
+<div class="cookies-component">
+  <div class="inner">
+    <div class="left">
+      <div class="logo"></div>
+      <div class="text">
+        CLARK uses cookies to provide with you with the best user experience. By using this platform you accept our use of cookies.
```
DonSeannelly: This should be placed into a strings file for easy updates/translation.

PR #436, opened 2018-10-04
Only show released to non-admins

DonSeannelly:
The browse page has been updated to only show objects that have been released to users that are not on the whitelist. Since users now can only browse non-released objects, the counter in the searchbar has also been updated.


src/app/cube/browse/browse.component.ts
```diff
@@ -38,7 +39,8 @@ export class BrowseComponent implements OnInit, OnDestroy {
     standardOutcomes: [],
     orderBy: undefined,
     sortType: undefined,
-    collection: ''
+    collection: '',
+    released: this.auth.group.value !== AUTH_GROUP.ADMIN ? true : undefined
```
nrwinner: Is there a reason this is undefined and not false?

PR #440, opened 2018-10-05
Separate togglePublish functionality into a publish and an unpublish function

nrwinner:


src/app/onion/core/learning-object.service.ts
```diff
@@ -136,6 +126,28 @@ export class LearningObjectService {
...
+      this.auth.user.username,
+      learningObject.name
+    );
+
+    return this.http
+      .patch(
+        route,
+        {
+          id: learningObject.id,
+          published: !learningObject.published
```
DonSeannelly: If this method is to only unpublish a learning object, should we use `published: false` in the patch request instead of `published: !learningObject.published`?

PR #468, opened 2018-10-23
Modal for Client Version Check

nvisal1:
We have encountered situations where our users were operating on the wrong version of the client. Users that were on the platform during an update were not notified of the changes unless they refreshed their page. This PR addresses that issue by adding a health check. The client will poll gateway every 10 minutes to ensure that the client version is up-to-date. If it is not, gateway will return a status code of 426 (upgrade required). This will trigger a popup modal to appear in the client. The modal will notify the user that they are using an out-dated version and will allow them to refresh the page.


src/app/core/auth.service.ts
```diff
@@ -71,6 +71,28 @@ export class AuthService {
...
+    return this.http
+      .get(environment.apiURL + '/clientversion/' + appVersion,
+      {
+        withCredentials: true,
+        responseType: 'text'
+      })
+      .toPromise()
+      .then(
+        (val) => {
+          return Promise.resolve();
```
DonSeannelly: Why are we choosing not to return the value returned from the API here?

src/app/core/auth.service.ts
```diff
@@ -71,6 +71,28 @@ export class AuthService {
     return this.user ? this.user.username : undefined;
   }
 
+  checkClientVersion(): Promise<void> {
+    // Application display name and Version information
```
DonSeannelly: Is the display name suppose to be included here?

PR #478, opened 2018-10-31
delete single object with top button

nvisal1:
Add specific check in the dashboard for deleting a single object with the delete multiple button


nrwinner:
Closes #506

PR #484, opened 2018-11-05
Details filesize

joelbutler1010:


src/app/shared/filesystem/file-list-view/file-list-view.component.scss
```diff
@@ -33,12 +33,12 @@
 
 .grid-row {
   display: grid;
-  grid-template-columns: 2fr 2fr 1fr 25px;
+  grid-template-columns: 2fr 2fr 2fr 1fr 25px;
```
nrwinner: The file size column shouldn't need 2fr width, should probably be 1fr so that it shrinks before the name and description fields

src/app/shared/filesystem/file-list-view/file-list-view.component.scss
```diff
@@ -33,12 +33,12 @@
...
   grid-column-gap: 20px;
   align-items: center;
   width: 100%;
   background: #fff;
-  padding: 0 8px 8px 0;
+  padding: 0 0 8px 0;
```
nrwinner: Why is the right-padding removed here?

src/app/shared/filesystem/file-list-view/file-size.pipe.ts
```diff
@@ -0,0 +1,38 @@
...
+  private units = [
+    'bytes',
+    'KB',
+    'MB',
+    'GB',
+    'TB',
+    'PB'
+  ];
+
+  transform(bytes: number = 0, precision: number = 2 ): string {
```
nrwinner: Thank you for this 😍

joelbutler1010: You know, just in case...

src/app/shared/filesystem/file-list-view/file-list-view.component.scss
joelbutler1010: If I didn't, we'd have this: 
sad

Instead of this:
happy

src/app/shared/filesystem/file-list-view/file-list-view.component.scss
joelbutler1010: Yup, I agree. 1fr is fine. I'll change it.

PR #499, opened 2018-11-15
Implement homepage connection banner and error state in learning object cards

nrwinner:


src/app/cube/browse/browse.component.ts
```diff
@@ -114,11 +116,18 @@ export class BrowseComponent implements OnInit, OnDestroy {
...
+        const collections = await this.collectionService.getCollections();
+        this.filters[0].values = collections.map(c => ({ name: c.name, value: c.abvName}));
+        this.query.released = this.auth.group.getValue() !== AUTH_GROUP.ADMIN ? true : undefined;
+        this.makeQuery(params);
+        this.fetchLearningObjects(this.query);
+      } catch (error) {
+        console.log(error);
+        this.loading = false;
+        this.error = 'There was an error retrieving this learning object!';
+        console.log(error);
```
DonSeannelly: Is there a reason for the logging here?

src/app/shared/learning-object/learning-object.component.ts
```diff
@@ -47,25 +48,12 @@ export class LearningObjectListingComponent implements OnInit, OnChanges {
 
     this.collectionService.getCollections().then(collections => {
       this.collections = new Map(collections.map(c => [c.abvName, c.name] as [string, string]));
+    }).catch(error => {
```
DonSeannelly: Is the intention of this catch statement to only suppress the error?

src/app/shared/learning-object/learning-object.component.ts
```diff
@@ -47,25 +48,12 @@ export class LearningObjectListingComponent implements OnInit, OnChanges {
...
     });
   }
 
   goals() {
     const punc = ['.', '!', '?'];
-    // convert goals to an array of strings containing the goal text
```
DonSeannelly: Why were the comments removed? This still seems to be an operation that isn't quite straightforward.

nrwinner: These comments were removed because most of the code here wasn't used. The deleted blob was a collection of unused code and comments pertaining to the unused code.

src/app/cube/browse/browse.component.ts
nrwinner: Nope. Fixed.

src/app/shared/learning-object/learning-object.component.ts
nrwinner: That was extraneous code. Removed.

PR #519, opened 2018-12-20
Use Access Groups instead of the Whitelist

DonSeannelly:
This PR eliminates the use of the whitelist in favor if the access groups that are listed on each user. All previously whitelisted users are now considered to be "reviewers."

PR #520, opened 2019-01-02
Docker e2e

nvisal1:
This PR includes necessary changes for e2e tests to be moved to a container orchestration. 
- Remove Cypress as a dev dependency 
- Create a separate Dockerfile for Cypress


src/environments/environment.ts
```diff
@@ -10,7 +10,8 @@ export const environment = {
   suggestionUrl: 'http://localhost:7000',
   contentManagerURL: 'http://localhost:5000',
   whiteListURL:
-    'https://raw.githubusercontent.com/Cyber4All/clark-client/master/whitelist/whitelist.json'
+    'https://raw.githubusercontent.com/Cyber4All/clark-client/master/whitelist/whitelist.json',
```
DonSeannelly: This will not be needed once #519 goes in.

PR #527, opened 2019-01-08
Created Edit Button on Action Panel

### LINKED ISSUE #524: Add Edit Button to Details Page
As a Collection Lead/Curator/Editor, I will be given an edit button on the details page so that I can quickly navigate to the page in which I can make edits to basic information and outcomes.

jbarlow96:
Resolves #524


src/app/cube/details/action-panel/action-panel.component.html
```diff
@@ -19,6 +19,11 @@
         <i class="fal fa-spinner-third fa-spin"></i>
       </span>
     </button>
+
+    <button (click)="clark.center" class="button neutral on-white" id="edit-button" type="submit"> 
+      {{ 'Revise' }}
```
DonSeannelly: Since this is not making a function call or computing any values, we could actually just directly write the text into the HTML instead of having Angular render it:
```suggestion
      Revise
```

src/app/cube/details/action-panel/action-panel.component.html
```diff
@@ -19,6 +19,11 @@
         <i class="fal fa-spinner-third fa-spin"></i>
       </span>
     </button>
+
+    <button (click)="clark.center" class="button neutral on-white" id="edit-button" type="submit">
```
DonSeannelly: The next step we'll need to take here is to only show this button if the user's AuthGroup is editor or admin.

src/app/cube/details/action-panel/action-panel.component.ts
```diff
@@ -251,6 +253,11 @@ export class ActionPanelComponent implements OnInit, OnDestroy {
     return this.windowWidth <= 750;
   }
 
+  openAdminApp() {
```
DonSeannelly: We should add a JSDoc comment here to describe what purpose the function has and what it does.

src/app/cube/details/action-panel/action-panel.component.ts
```diff
@@ -45,6 +46,8 @@ export class ActionPanelComponent implements OnInit, OnDestroy {
   windowWidth: number;
   loggedin = false;
   showDownloadModal = false;
+  authGroup: AUTH_GROUP = 0;
```
DonSeannelly: Is this being used in this component? What was the reason for adding it here?

src/app/cube/details/action-panel/action-panel.component.html
```diff
@@ -19,6 +19,19 @@
...
       </span>
     </button>
+
+    <span class="btn-group vertical" *ngIf="isViewable">
+      <button class="button neutral on-white" id="edit-button" type="submit"> 
+        <a (click)="openAdminApp()">Revise</a>
+
+      </button>
+    </span>
+    <!--
```
DonSeannelly: We shouldn't leave commented out code lingering in the code base. If it isn't being used we should delete it to reduce clutter.

src/app/core/auth.service.ts
```diff
@@ -314,6 +315,12 @@ export class AuthService {
     return this.group.getValue() > AUTH_GROUP.USER;
   }
 
+  /**
+   * Identifies if the user should be able to see the "revise" button on the action panel of a learning object.
```
DonSeannelly: Instead of commenting this function from the perspective of where it is used at the moment, we should document it in a way that would still make sense if used for another reason. This also applies to the name of the function. Perhaps something like `hasEditorPrivilege` would abstract it away from the action panel a bit better?

PR #535, opened 2019-01-14
Fix review pending banner not showing

DonSeannelly:
Due to a recent change in the authorization logic, the review pending banner no longer shows for Learning Objects that are not released. This is due to the fact that the `isReleased` variable was being set based on the user's privilege level and not the Learning Object's status.

This PR updates the component logic related to download access and release status.

### DIFF
--- a/src/app/cube/details/action-panel/action-panel.component.ts
+++ b/src/app/cube/details/action-panel/action-panel.component.ts
@@ -1,14 +1,10 @@
-import { CartV2Service, iframeParentID } from '../../../core/cartv2.service';
+import { Component, ElementRef, HostListener, Input, OnDestroy, OnInit, Renderer2, ViewChild } from '@angular/core';
 import { LearningObject, User } from '@cyber4all/clark-entity';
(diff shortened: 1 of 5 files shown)


src/app/cube/details/action-panel/action-panel.component.ts
```diff
@@ -73,25 +69,16 @@ export class ActionPanelComponent implements OnInit, OnDestroy {
...
     this.userIsAuthor = (this.learningObject.author.username === userName);
   }
 
-  get canDownload(): boolean {
-    return this.downloadStatus === DOWNLOAD_STATUS.CAN_DOWNLOAD;
-  }
-
   get isReleased(): boolean {
-    return this.downloadStatus !== DOWNLOAD_STATUS.NOT_RELEASED;
+    return this.learningObject['status'] === 'released';
```
nrwinner:
```suggestion
    // FIXME status should be accessed as learningObject.status and should be compared to the enum in clark-entity
    return this.learningObject['status'] === 'released';
```

PR #540, opened 2019-01-16
Fix revise button showing after logout

DonSeannelly:
When an admin or editor logged out of the app while on the details page, the revise button remained visible on the page. This PR updates the Action Panel to check the user's access privileges each time the authorization group changes in the AuthService.


src/app/cube/details/action-panel/action-panel.component.ts
```diff
@@ -68,12 +69,15 @@ export class ActionPanelComponent implements OnInit, OnDestroy {
     private cartService: CartV2Service,
     private renderer: Renderer2,
     private noteService: ToasterService,
-  ) {}
+  ) { }
 
   ngOnInit() {
+    this.auth.group
+      .takeUntil(this.isDestroyed$)
```
nrwinner:
```suggestion
      .pipe(takeUntil(this.isDestroyed$))
```

PR #553, opened 2019-01-18
Modified makeQuery to check for filter values and update filters

### LINKED ISSUE #510: Collection filter disappears after search
When selecting a collection, adding a search term, and the results are loaded, the filter for the collection disappears and must be re-selected to continue searching by that specific collection.  For instance, click security injections as the filter and type "buffer overflow".

EVogel1999:
Fixes issue where filter values do not reset on text search.  Does this by adding a flag in makeQuery() which takes the params in the url and checks for the three main filter key values (length, level, and collection).  If none of the filter values are in the params it resets the filter variable by using the clearAllFilters() function.
Closes #510


src/app/cube/browse/browse.component.ts
```diff
@@ -353,10 +353,13 @@ export class BrowseComponent implements OnInit, OnDestroy {
    */
   makeQuery(params: any) {
     const paramKeys = Object.keys(params);
+    let filters = false;
     // iterate params object
     for (let i = 0, l = paramKeys.length; i < l; i++) {
       const key = paramKeys[i];
-
+      if (key === 'length' || key === 'level' || key === 'collection') {
```
nrwinner: Why did we hardcode these values here?

Issue #563, opened 2019-01-24
Add the collection disclaimer to objects details page and tooltip to bar


siddharthkaza:
On the top bar - we should add a tooltip or "?" (in attached screenshot, next to NSA NCCP) that points to the collection page: https://www.clark.center/c/NSA%20NCCP 

In addition, we need a new area (possibly above reviews) or a "notes" area which lists the disclaimer for the collection. Once disclaimer is added, remove the disclaimer from page footer.

DonSeannelly:
@siddharthkaza you mention an attached screenshot that I am not seeing here. Should there be an image in this thread?

pzalep1:
![image](https://user-images.githubusercontent.com/43140629/51758633-1faa1680-2094-11e9-9acd-e07396454682.png)

@siddharthkaza @DonSeannelly Is this where the tooltip should go and should look like? Also, would this tooltip be present on all learning objects in every collection and link to their corresponding collection pages or is this just a feature for NCCP learning objects?

PR #569, opened 2019-01-25
Change disclaimer location on NCCP learning object pages

pzalep1:
Move disclaimer location to NCCP learning object pages and out of footer. Located between materials and ratings on those pages. 
Contributes to #563


DonSeannelly:
@nrwinner can you check the styling on this?

PR #577, opened 2019-01-28
Fix bug where DirectoryTree attempts to delete from a null node

nrwinner:
This PR addresses a bug wherein after deleting the contents of the root directory, the DirectoryTree attempts to delete the root directory (since it's empty) and throws a javascript error, preventing further interaction with the FileBrowser.

### DIFF
--- a/src/app/shared/filesystem/DirectoryTree.ts
+++ b/src/app/shared/filesystem/DirectoryTree.ts
@@ -238,7 +238,7 @@ export class DirectoryTree {
    * @memberof DirectoryTree
    */
   private removeEmptyFolders(node: DirectoryNode): void {
-    if (!node.getChildren().length && !node.getFiles().length) {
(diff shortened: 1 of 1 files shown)


src/app/shared/filesystem/DirectoryTree.ts
```diff
@@ -238,7 +238,7 @@ export class DirectoryTree {
    * @memberof DirectoryTree
    */
   private removeEmptyFolders(node: DirectoryNode): void {
-    if (!node.getChildren().length && !node.getFiles().length) {
+    if (!node && !node.getChildren().length && !node.getFiles().length) {
```
gshaw1997: @nrwinner The first check here checks if `node` is falsy. If `node` is falsy, wouldn't that cause the subsequent checks and code within the if block to fail. Am I missing something?

nrwinner:
Included in #588

PR #581, opened 2019-01-29
Sort verbs in dropdown for learning outcomes

nrwinner:
This PR implements verb sorting in the builder. Ideally, the verbs would not be sorted client side and this PR should be reverted when a more permanent solution is found.


nrwinner:
> Why not update them here instead: https://github.com/Cyber4All/clark-taxonomy/blob/master/lib/taxonomy.ts?

 I agree that's a better way to do it.

PR #593, opened 2019-01-31
Update to Angular 7

pzalep1:
Updated to Angular 7


src/main.ts
```diff
@@ -5,9 +5,9 @@ import { ClarkModule } from 'app/clark.module';
 import { environment } from '@env/environment';
 
 // Global rxjs imports
```
nrwinner: Can we remove this comment if we're not importing anything under it?

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compileOnSave": false,
   "compilerOptions": {
+    "importHelpers": true,
```
nrwinner: What is this?

src/main.ts
pzalep1: I missed this when I was looking through. I went back and removed it.

tsconfig.json
pzalep1: image

PR #598, opened 2019-02-06
Force page reload on logout

nrwinner:
This PR implements a page reload on user logout. This helps to ensure that all user data from the previously logged-in user is scrubbed from the system and that relevant security properties such as access groups don't linger.

### DIFF
--- a/src/app/shared/navbar/navbar.component.ts
+++ b/src/app/shared/navbar/navbar.component.ts
@@ -157,7 +157,9 @@ export class NavbarComponent implements OnInit, AfterContentChecked, OnDestroy {
   }
 
   logout() {
-    this.authService.logout();
+    this.authService.logout().then(() => {
+      window.location.reload();
+    });
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/app/shared/navbar/navbar.component.ts
```diff
@@ -157,7 +157,9 @@ export class NavbarComponent implements OnInit, AfterContentChecked, OnDestroy {
   }
 
   logout() {
-    this.authService.logout();
+    this.authService.logout().then(() => {
+      window.location.reload();
```
DonSeannelly: What's the reason for doing this in the component instead of the service?

nrwinner: I didn't personally think that the service should have any knowledge of the actions taken after logging out, that it should instead only know to query the service and update state. Our reload seems to me to be concern external to the service. Thoughts?

DonSeannelly: @nrwinner I was viewing it a little differently. I was considering refreshing the page the last step of the logout process, not something that the navbar does because the user logged out.

Is this refresh something unique to the navbar functionality? Or if we had a logout button somewhere else, would that also refresh the app?

nrwinner: I'm sold. Making the change.

PR #620, opened 2019-02-13
List contributors

### LINKED ISSUE #612: List all contributors on search result cards
List all contributors (and creator) separated by commas. To save space, we can remove organizations. 

screen shot 2019-02-08 at 1 23 23 pm

Ballen2713:
pic2
pic1

Closes #612

### DIFF
--- a/src/app/shared/learning-object/learning-object.component.html
+++ b/src/app/shared/learning-object/learning-object.component.html
@@ -19,7 +19,10 @@
           <span *ngFor="let i of learningObject.contributors.slice(0,2)">
             , {{ i.name }}
           </span>
-          <span *ngIf="learningObject.contributors.length-2 > 0 "><span tip="{{ getContributors() }}">, +{{ learningObject.contributors.length - 2 }} more</span></span>
+          <span *ngIf="learningObject.contributors.length-2 > 0 ">
+            <span class="toolTip" tip="{{ getContributors() }}" >, and {{ learningObject.contributors.length - 2 }} other<span *ngIf="learningObject.contributors.length-2 > 2">s</span>
+          </span>
+        </span>
         </span>
       </div>
       <div class="date">
--- a/src/app/shared/learning-object/learning-object.component.html
+++ b/src/app/shared/learning-object/learning-object.component.html
@@ -14,7 +14,12 @@
     </div>
     <div class="details">
       <div class="author">
-        {{ learningObject.author.name | titlecase }} | {{ learningObject.author.organization }}
+        {{ learningObject.author.name | titlecase }} 
+        <span *ngFor="let i of learningObject.contributors">
+          <span *ngIf="learningObject.contributors.length > 2" tip="{{ i.name | titlecase }}">
+            , {{ i.name }}
+          </span> 
+        </span>
       </div>
       <div class="date">
         Updated {{ date | date:'mediumDate' }}
--- a/src/app/shared/learning-object/learning-object.component.html
+++ b/src/app/shared/learning-object/learning-object.component.html
@@ -15,11 +15,12 @@
     <div class="details">
       <div class="author">
         {{ learningObject.author.name | titlecase }}
-        <span *ngFor="let i of learningObject.contributors">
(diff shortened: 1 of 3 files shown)


nrwinner:
Looks good! The contributors should read a little better though, for example the the names in the tooltip read "name1 ,name2" instead of "name1, name2". Also, if we have more than 2, ideally the text should read "name1, name2, and 3 others".

src/app/shared/learning-object/learning-object.component.scss
```diff
@@ -1,6 +1,9 @@
 @import '~_vars.scss';
 
 @media (min-width: 751px) {
+    .toolTip { 
+        text-transform: lowercase;
```
nrwinner: Why are we displaying text content lowercased inside the tooltip? Also, why is that logic dependent on the size of the screen?

src/app/shared/learning-object/learning-object.component.html
```diff
@@ -19,7 +19,10 @@
           <span *ngFor="let i of learningObject.contributors.slice(0,2)">
             , {{ i.name }}
           </span>
-          <span *ngIf="learningObject.contributors.length-2 > 0 "><span tip="{{ getContributors() }}">, +{{ learningObject.contributors.length - 2 }} more</span></span>
+          <span *ngIf="learningObject.contributors.length-2 > 0 ">
+            <span class="toolTip" tip="{{ getContributors() }}" >, and {{ learningObject.contributors.length - 2 }} other<span *ngIf="learningObject.contributors.length-2 > 2">s</span>
```
nrwinner: `<span *ngIf="learningObject.contributors.length-2 > 2">s</span>` means that the s will only render if we have 3 or more, 2 or more won't render it. For example `[user1, user2]` has 2 users, but since 2 !> 2, we'd get `and 2 other`

src/app/shared/learning-object/learning-object.component.ts
```diff
@@ -114,14 +114,14 @@ export class LearningObjectListingComponent implements OnInit, OnChanges {
   //Method for title case inside tool tip
   toTitleCase = function (str) {
```
nrwinner: Is there any reason we're not using the Angular `titleCase` pipe or just CSS to render this?

src/app/shared/learning-object/learning-object.component.ts
```diff
@@ -114,14 +114,14 @@ export class LearningObjectListingComponent implements OnInit, OnChanges {
...
-    for (var i = 0; i < str.length; i++) {
+    for (let i = 0; i < str.length; i++) {
       str[i] = str[i].charAt(0).toUpperCase() + str[i].slice(1);
     }
     return str.join(' ');
   };
 
   getContributors() { 
-    return this.learningObject.contributors.map(x => this.toTitleCase(x.name + ' '));
+    return this.learningObject.contributors.map(x => this.toTitleCase(' ' + x.name));
```
nrwinner: Won't this add extra spaces between names? Additionally, adding the space at index 0 of the string, means the logic in the above function (`str[I].charAt(0)`) will attempt to capitalize the space instead of capitalizing the first letter of the word.

Ballen2713: Not necessarily, because the extra space I concatenated with the name is to change the position of the comm

a. Without the space added the names of the contributors will be squished together like so and you can see that the title case method still works the same way withoutspace

src/app/shared/learning-object/learning-object.component.ts
Ballen2713: Yes using the pipe titlecase breaks the UI and I don't see a way to use css to capitalize whats inside the tooltip. I will include pictures of the error and what happens if I use the titlecase pipe 
usingtitlecase1
usingtitlecase

src/app/shared/learning-object/learning-object.component.scss
Ballen2713: I didn't mean to put that inside the @media scope but this is how I am able to lowercase the 'and' and 'other' words sense they were automatically being converted to titlecase

nrwinner:
Closed in #658

PR #640, opened 2019-02-19
Hide unreleased children on details page

nvisal1:
Filter children on details page to hide unreleased objects


src/app/cube/details/details.component.ts
```diff
@@ -109,6 +110,14 @@ export class DetailsComponent implements OnInit, OnDestroy {
           return file;
         }
       );
+      this.children = this.learningObject.children.filter(
```
DonSeannelly: Can we add a FIXME to this, stating that it should be removed once the service logic is updated? It's just a temporary addition

PR #646, opened 2019-02-20
Fix bad refresh token request

### LINKED ISSUE #607: Client Should Handle Invalid Access Tokens
Sentry Issue: [CLARK-CLIENT-8R](https://sentry.io/cyber4all/clark-client/issues/847304055/?referrer=github_integration)

```
Error: Uncaught (in promise): e: {"headers":{"normalizedNames":{},"lazyUpdate":null},"status":401,"statusText":"OK","url":"https://api-gateway.clark.center/users/tokens/refresh","ok":false,"name":"HttpErrorResponse","message":"Http failure response for https://api-gateway.clark.center/users/tokens/refresh: 401 OK","error":"Invalid Access Token"}
  at j (/polyfills.db8a145c55ae59b50492.js:1:13230)
  at None (/polyfills.db8a145c55ae59b50492.js:1:12330)
  at a (/main.94e92f40d40dea32455d.js:1:956001)
  at e.invoke (/polyfills.db8a145c55ae59b50492.js:1:7288)
  at Object.onInvoke (/main.94e92f40d40dea32455d.js:1:257296)
...
(11 additional frame(s) were not displayed)
```

DonSeannelly:
The client app was making a request to refresh the user's token even when the token was invalid. Now the app first validates the token before it attempts to refresh it. The refresh operation only happens if the token is valid. If the token is not valid, it will prompt the user to sign in again.

This preserves the operation of automatically updating the user information stored in the token if the token is valid, while adding appropriate handling if it is not valid.

While the request is made, a loading spinner is shown to the user:

image

Closes #607

### DIFF
--- a/src/app/auth/email-verified/email-verified.component.ts
+++ b/src/app/auth/email-verified/email-verified.component.ts
@@ -5,7 +5,7 @@ import { AuthService } from '../../core/auth.service';
   selector: 'app-email-verified',
   styles: [
     `.message--in-progress {
-      color: red;
+      margin-bottom: 20px;
     }`
   ],
   templateUrl: './email-verified.component.html'
--- a/src/app/auth/email-verified/email-verified.component.ts
+++ b/src/app/auth/email-verified/email-verified.component.ts
@@ -3,12 +3,31 @@ import { AuthService } from '../../core/auth.service';
 
 @Component({
   selector: 'app-email-verified',
+  styles: [
+    `.message--in-progress {
+      color: red;
+    }`
+  ],
   templateUrl: './email-verified.component.html'
 })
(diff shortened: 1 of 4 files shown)


src/app/auth/email-verified/email-verified.component.ts
```diff
@@ -3,12 +3,31 @@ import { AuthService } from '../../core/auth.service';
 
 @Component({
   selector: 'app-email-verified',
+  styles: [
+    `.message--in-progress {
```
nrwinner: yasssss

DonSeannelly:
Updated screen shot, with margin applied:

image

PR #649, opened 2019-02-20
Fix bug where users with privileged access can't download learning objects

nrwinner:
**Problem:** With the addition of collections into the access groups (`reviewer` changing to `reviewer@collection`) the current implementation client side identified reviewers and curators as regular users and this denied them download access.

**Solution:** Before iterating the accessGroups array, remove references to collections entirely.

**Future:** A more robust solution should be implemented wherein there are collection checks client side to prevent a user downloading an object from a collection they're not privileged in. This isn't absolutely necessary at this time since the service will only send down objects that the current user is authorized to see. @gshaw1997 @davidnso can you confirm this last bit?

### DIFF
--- a/src/app/core/auth.service.ts
+++ b/src/app/core/auth.service.ts
@@ -16,6 +16,7 @@ export enum AUTH_GROUP {
   VISITOR,
   USER,
   REVIEWER,
+  CURATOR,
   EDITOR,
   ADMIN
 }
@@ -299,14 +300,14 @@ export class AuthService {
   }
 
   /**
-   * Identifies if the current logged in user has reviewer priviledges.
+   * Identifies if the current logged in user has reviewer privileges.
    */
   hasReviewerAccess(): boolean {
     return this.group.getValue() > AUTH_GROUP.USER;
   }
 
   /**
-   * Identifies if the current logged in user has editor priviledges.
+   * Identifies if the current logged in user has editor privileges.
    */
   public hasEditorAccess(): boolean {
     return this.group.getValue() > AUTH_GROUP.REVIEWER;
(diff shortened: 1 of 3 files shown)


src/app/core/auth.service.ts
```diff
@@ -299,14 +300,14 @@ export class AuthService {
   }
 
   /**
-   * Identifies if the current logged in user has reviewer priviledges.
+   * Identifies if the current logged in user has reviewer privileges.
```
DonSeannelly: ❤️

src/app/core/auth.service.ts
```diff
@@ -319,12 +320,17 @@ export class AuthService {
     if (!this.user['accessGroups']) {
       return;
     }
-    if (this.user['accessGroups'].includes('admin')) {
+
+    const groups = this.user['accessGroups'].map(x => x.split('@')[0]);
```
DonSeannelly: If this is the solution we want to go with, can we add a comment here to explain this as it was explained in the PR description?

src/app/core/auth.service.ts
nrwinner: The funny part is I misspelled the same word in my commit message

gshaw1997:
> **Problem:** With the addition of collections into the access groups (`reviewer` changing to `reviewer@collection`) the current implementation client side identified reviewers and curators as regular users and this denied them download access.
> 
> **Solution:** Before iterating the accessGroups array, remove references to collections entirely.
> 
> **Future:** A more robust solution should be implemented wherein there are collection checks client side to prevent a user downloading an object from a collection they're not privileged in. This isn't absolutely necessary at this time since the service will only send down objects that the current user is authorized to see. @gshaw1997 @davidnso can you confirm this last bit?

@nrwinner Yes, they will only see objects they have access to.

PR #675, opened 2019-02-25
Fix failing tests

DonSeannelly:
### Goal
Unit tests are failing across the app due to the TestBed for components not being set up properly. This PR begins to fix this issue.

### Tests Removed
- AuthComponent, there is no logic here so tests were only making sure the module could initialize. We can cover this effectively with e2e tests

### CI Updates
CI will now run tests in the AuthModule, as all tests there are working. As we fix more tests in the app, this should eventually expand to run all tests.


.circleci/config.yml
```diff
@@ -24,6 +24,9 @@ jobs:
     steps:  
       - attach_workspace : 
           at: ~/clark-client
+      - run:
```
nrwinner: Yasssss :heart:

.circleci/config.yml
```diff
@@ -12,6 +12,9 @@ jobs:
       - run:
           name: Install Dependencies
           command: npm install
+      - run:
```
nrwinner: Oh nvm yasssssss :heart:

Issue #669, opened 2019-02-22
Search for learning objects in the curator dashboard


nrwinner:
As a curator/editor/admin I want to be able to search for learning objects I have access to so that I can facilitate the review processes.

Needs #667

PR #684, opened 2019-02-28
Search for Learning Objects in the Curator Dashboard (Unit 2 of 2)

nvisal1:
Deployable unit 2 includes: 
- Implement learning object search functionality 
- Implement virtual scroll for learning object list 
- Add loading spinner when searching for learning objects
- Add admin context menu functionality (view user, edit, change status)
- Use clark-popup for change status modal

This PR has a lot of files changed because I needed to merge master to use the new learning object save function for changing status

Closes https://github.com/Cyber4All/clark-client/issues/669

Screenshot
screen shot 2019-02-28 at 8 43 36 am

Updated Screenshot
screen shot 2019-02-28 at 1 20 49 pm


src/app/admin/components/search-bar/search-bar.component.ts
```diff
@@ -1,15 +1,29 @@
...
+  searchInput: ElementRef;
+  destroyed$: Subject<void> = new Subject();
 
   constructor() { }
 
-  ngOnInit() {
+  ngAfterViewInit() {
+    fromEvent(this.searchInput.nativeElement, 'input').pipe(
+      map(x => x['currentTarget'].value),
+      debounceTime(650),
```
DonSeannelly: Should `debounceTime` happen first so that we don't map unless the timer has passed?

src/app/admin/pages/learning-objects/learning-objects.component.html
```diff
@@ -1,11 +1,14 @@
 <clark-admin-content-wrapper>
   <div page-title>Learning Objects</div>
-  <clark-admin-search-bar search></clark-admin-search-bar>
+  <clark-admin-search-bar search 
+    [placeholder]="searchBarPlaceholder"
+    (userInput)="getLearningObjects($event)"
+  ></clark-admin-search-bar>
   <div filter-options>filter options yo</div>
 
   <div class="list" main>
```
DonSeannelly: This is a nitpick, but I think it might be helpful to take the whole list div and move it into its own component. It would bring this components template size down a bit, and create a better separation of concerns by leaving one component to compose the page and another to only take charge of displaying the list. I'm interested in other's thoughts on this as well.

src/app/admin/pages/learning-objects/learning-objects.component.ts
```diff
@@ -1,69 +1,71 @@
 import { Component, OnInit } from '@angular/core';
+import { LearningObjectService as publicService } from 'app/cube/learning-object.service';
```
DonSeannelly: Overall the fact that we are needing to alias like this smells of bad design, however I think a refactoring here might be out of scope for this PR. I do however think that the alias names should:
1. Be descriptive of the service type, not just generically a "service"
2. Maintain the casing of the aliased member, instead of converting to camelCase

src/app/admin/pages/learning-objects/learning-objects.component.ts
```diff
@@ -1,69 +1,71 @@
...
-        published: true,
-        name: 'Integer Error - CS2 - C++',
-        length: 'nanomodule',
-        materials:{
-          files:[],
-          urls:[
-            // tslint:disable-next-line:max-line-length
-            {title:'Online module','url':'http://cis1.towson.edu/~cyber4all/modules/nanomodules/Integer_Error-CS2_C++.html'},{'title':'See a Security Injection Module in Action','url':'https://youtu.be/MH_RD1jh0AE'},{'title':'How to use this module in your class','url':'https://youtu.be/u47q-qX52JI'}],'notes':'','pdf':{'name':'0ReadMeFirst - Integer Error - CS2 - C++.pdf','url':'https://neutrino-file-uploads.s3.us-east-2.amazonaws.com/skaza/5aa0183becba9a264dcd806b/0ReadMeFirst%20-%20Integer%20Error%20-%20CS2%20-%20C%2B%2B.pdf'}},'levels':['undergraduate'],'contributors':['5c6d71c29554c723fba68b19'],'collection':'secinj','lock':{'date':{'$numberDouble':'1538061211867'},'restrictions':['download']},'status':'released','description':'Introduce the concept of integer overflow in first programming courses'},
-      ];
+  ngOnInit() {}
```
DonSeannelly: If this lifecycle hook is unused we should remove it and the declaration of the OnInit interface implementation.

src/app/admin/components/search-bar/search-bar.component.ts
nvisal1: @nrwinner any thought on this? I tried playing around with this and found that the currentTarget property is set to null after debounce.

nrwinner: I'm not sure, I wouldn't think currentTarget would be null after the debounce unless the event is somehow expiring?

PR #688, opened 2019-02-28
Implement routing changes to allow for different admin/curator views

nrwinner:


src/app/admin/admin.component.html
```diff
@@ -1,8 +1,13 @@
 <div class="admin">
   <div class="admin__sidebar">
-    <clark-admin-sidebar></clark-admin-sidebar>
+    <clark-admin-sidebar
+      [authorizedCollections]="authorizedCollections"
```
DonSeannelly: If the authorization of what collections should be seen is happening outside of this component, could we just call this property collections and dumb down the sidebar component to not even know the difference?

src/app/admin/admin.component.ts
```diff
@@ -1,21 +1,94 @@
...
+  get initialized(): boolean {
+    return !this._initialized.includes(false);
+  }
 
   ngOnInit() {
+    // hide CLARK navbar
     this.navbarService.hide();
+
+    if (this.authService.hasEditorAccess()) {
+      this.adminMode = true;
```
DonSeannelly: Why do we enable admin mode for an editor?

src/app/admin/admin.component.ts
```diff
@@ -1,21 +1,94 @@
...
+    return await this.authService.isLoggedIn.pipe(
+      skipWhile(x => x === false),
+      take(1)
+    ).subscribe(async () => {
+      // we're sure the user is logged in here and so access groups should be defined
+      return await Promise.all(this.authService.user['accessGroups']
+        .filter(group => group.includes('curator@'))
+        .map(group =>
+          this.collectionService.getCollection(group.split('@')[1]).then(c => this.authorizedCollections.push(c))
+        )).then(() => {
```
DonSeannelly: Can we drop the `.then` to a new line for consistency sake?

src/app/admin/components/sidebar/sidebar.component.ts
```diff
@@ -1,25 +1,39 @@
...
   selector: 'clark-admin-sidebar',
   templateUrl: './sidebar.component.html',
   styleUrls: ['./sidebar.component.scss']
 })
-export class SidebarComponent implements OnInit {
+export class SidebarComponent implements OnInit, OnDestroy {
+  destroyed$: Subject<void> = new Subject();
 
-  constructor(private location: Location) { }
+  @Input() authorizedCollections: string[] = [];
```
DonSeannelly: See above comment, but I think this would be better named as `collections` since the component is not dealing with anything more than the collections it is given.

src/app/core/auth.service.ts
```diff
@@ -363,14 +363,14 @@ export class AuthService {
    * @memberof AuthService
    */
   public hasCuratorAccess(): boolean {
-    return this.group.getValue() > AUTH_GROUP.CURATOR;
+    return this.group.getValue() > AUTH_GROUP.REVIEWER;
```
DonSeannelly: Good catch 👍

PR #690, opened 2019-03-01
Fix files not appearing on details page when navigating to a Learning Object that has files with the same path

### LINKED ISSUE #641: Linked learning object material partially or fully not appearing until after refresh
Observed some learning object (that are in a linked relationship) material may not appear partially or fully until after the page is refreshed. This bug was first seen in with the course "National Security, Electronic Surveillance, and the Fourth Amendment". 

before_refresh
after_refresh

gshaw1997:
This PR fixes an issue in the file system where files with different `id`s, but the same path added back to back would not appear in the file browser. 

This occurred because the file system performed its clean operation after adding nodes to clear cached file references and remove files that should no longer exist. When a new array of files are added, the file system would compare the new files against what it has in its cache by the files' ids. This would trigger a delete operation of that file at the path stored in the cache. When the path of the new file and cached file were the same, the file system would end up removing the new file while attempting to remove the cached file (b/c paths are unique in the file system's context).

This was fixed by comparing both id and path before removing.

Closes #641

### DIFF
--- a/src/app/shared/filesystem/DirectoryTree.ts
+++ b/src/app/shared/filesystem/DirectoryTree.ts
@@ -34,7 +34,6 @@ export class DirectoryTree {
         }
         return true;
       });
-
     const newFiles = getUpdatedFiles(files);
     for (const file of newFiles) {
       this.fileMap.set(file.id, {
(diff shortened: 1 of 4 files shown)


src/app/shared/filesystem/DirectoryTree.ts
```diff
@@ -65,10 +64,36 @@ export class DirectoryTree {
...
    */
   private cleanFilesystem(files: LearningObject.Material.File[]) {
-    const fileIds = files.map(file => file.id);
+    /**
+     * Checks if file is in array by matching against id and path
+     *
+     * @param {{
+     *       cachedFileData: { id: string; path: string };
+     *       file: LearningObject.Material.File
+     *     }} params
```
DonSeannelly: I think that this is auto-generated JSDoc from a tool. I'd vote that we either remove it, or take some time to write doc that explains these properties. Currently this doesn't do anything that TypeScript doesn't already do.

I've seen it a lot throughout our code and I think it just clutters things up.

PR #691, opened 2019-03-01
Implement sidebar component in collection dashboard

nrwinner:
This PR implements the new sidebar for the collection dashboard

- Admins will see the correct three options (analytics, objects, users) across all collections
screen shot 2019-03-01 at 1 34 50 pm

- Curators will see a list all collections they curate for with a dropdown option.
screen shot 2019-03-01 at 1 31 39 pm

Branched off of #688 so that should be merged first. First new commit is 8ddff72

### DIFF
--- a/src/app/admin/admin.component.ts
+++ b/src/app/admin/admin.component.ts
@@ -20,7 +20,7 @@ export class AdminComponent implements OnInit, OnDestroy {
 
   adminMode: boolean;
 
-  private _initialized: boolean[] = [false];
+  private _initialized: boolean[] = [ false /* collections loaded */ ];
 
   constructor(
     private navbarService: NavbarService,
@@ -30,6 +30,13 @@ export class AdminComponent implements OnInit, OnDestroy {
     private collectionService: CollectionService
   ) {}
 
+  /**
+   * Returns true if the component is fully initialized (ie the array doesn't contain false), false otherwise
+   *
+   * @readonly
+   * @type {boolean}
+   * @memberof AdminComponent
+   */
   get initialized(): boolean {
     return !this._initialized.includes(false);
   }
@@ -40,8 +47,10 @@ export class AdminComponent implements OnInit, OnDestroy {
 
     if (this.authService.hasEditorAccess()) {
       this.adminMode = true;
+      // we don't need to load collections, so we can set that initialization block to true
+      this._initialized[0] = true;
     } else {
-      // fetch the route parameter representing the selected collection
+      // this user isn't an admin/editor so fetch the route parameter representing the selected collection
       this.route.paramMap
         .pipe(takeUntil(this.destroyed$))
         .subscribe(params => {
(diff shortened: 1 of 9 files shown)


src/app/admin/admin.component.ts
```diff
@@ -1,21 +1,98 @@
...
 
 @Component({
   selector: 'clark-admin',
   templateUrl: './admin.component.html',
   styleUrls: ['./admin.component.scss']
 })
 export class AdminComponent implements OnInit, OnDestroy {
+  destroyed$: Subject<void> = new Subject();
+
+  authorizedCollections: Collection[] = [];
```
DonSeannelly: Doesn't this variable name imply that there are unauthorized collections that this component is in some way concerned about?

src/app/admin/admin.component.ts
```diff
@@ -1,21 +1,98 @@
...
+   * Retrieve a list of collection names for which the user is authorized to view (has curator status for)
+   *
+   * @memberof AdminComponent
+   */
+  async retrieveAuthorizedCollections() {
+    // wait for user to be logged in (edge case) and then fetch list of curated collections
+    return await this.authService.isLoggedIn.pipe(
+      skipWhile(x => x === false),
+      take(1)
+    ).toPromise().then(async () => {
```
DonSeannelly: Can we drop the function chains to a new line to make the code more easy to scan?

src/app/admin/components/sidebar/sidebar.component.html
```diff
@@ -5,12 +5,30 @@
...
-      </ul>
+    <div *ngIf="initialized" class="sidebar__menu">
+      <ng-container *ngTemplateOutlet="!editorMode ? collectionMenu : sidebarMenu"></ng-container>
     </div>
   </div>
-</div>
\ No newline at end of file
+</div>
+
+<ng-template #sidebarMenu>
```
DonSeannelly: This template is the sidebar menu exclusively for admins, right?

src/app/admin/components/sidebar/sidebar.component.ts
```diff
@@ -1,25 +1,54 @@
...
+  @Input() activeCollection: string;
+  @Input() editorMode: boolean;
 
-  ngOnInit() {
-  }
+  @Input() initialized = false;
+
+  constructor(private router: Router) {}
+
+  ngOnInit() {}
```
DonSeannelly: If we aren't using this lifecycle hook we should remove it and the implementation of the interface.

src/app/admin/components/sidebar/sidebar.component.html
nrwinner: No, this template is the markup for menu options since it gets reused in two different places.

DonSeannelly: Ah, I got it now. Thanks!

src/app/admin/admin.component.ts
nrwinner: I would argue that, since this is the root level component and it is responsible for retrieving the collections that the currently logged in user is authorized to view in this context, this variable name makes sense?

DonSeannelly: Yeah, I can get behind that.

I think my uneasiness is perhaps around the fact that this is the `AdminComponent` yet it is being used by both admin and curator user types. While curators are only authorized for a subset of collections, admins should always be authorized for all collections. How do you feel about renaming this component?

You mentioned it is the root level component, would that make CollectionDashboardComponent a better name?

nrwinner: Yeah that probably makes more sense. What do you think we should do about the route? This currently lives at /admin, I feel collection-dashboard is a bit verbose?

Issue #632, opened 2019-02-15
Add collections to stats on home page and stats page


siddharthkaza:
As a visitor, I want to be able to see the number of curated collections so that I can better understand the platform.

------

Add "# Curated Collections" to the counter bar on homepage. 

<img width="632" alt="screen shot 2019-02-15 at 11 22 43 am" src="https://user-images.githubusercontent.com/2423261/52869646-24eb0600-3114-11e9-825d-b629a9d42ab7.png">

PR #701, opened 2019-03-08
Implement homepage redesign

nrwinner:
**This pull request implements a new homepage. The driving forces behind the redesign are:**

1) The lack of consistency between the onion themes and the cube themes
2) The absence of room for the new collection statistics
3) The absence of a description of a learning object and the disjointed ordering of content on the page

**This PR:**

1) Reorders the content on the page into a more thoughtful and explanatory flow
2) Refreshes the UI and componentizes much of the homepage
3) Introduces a new stats section to better highlight key metrics of CLARK 

**Future Endeavors** 

1) The philosophy section should no-longer show bullet points without context. Currently favored solution is a grid layout with clickable cards revealing more information for a particular philosophy point.
2) The _Learning Objects_ section should show a carousel to prevent the objects from awkwardly wrapping onto the next line while still providing a decent sampling of content.
3) Should a larger number of collections need to be featured, we should be able to reuse the carousel idea for them as well.

**Screenshots**
screencapture-localhost-4200-home-2019-03-11-12_25_15

Contributes to #632


nrwinner:
This PR is still unfinished but I would appreciate some design feedback. The code that's here can also be reviewed now (I think) in order to reduce review process of the final PR.

nvisal1:
This is awesome! I think that the top of the footer should be slanted for consistency. Thoughts?

nrwinner:
> This is awesome! I think that the top of the footer should be slanted for consistency. Thoughts?

@nvisal1 my plan there is, since there's still the usage stats and the call to action left to add, we just won't slant the call to action at all and kinda kill the pattern on the way down. I'm hesitant to slant the footer because the footer exists on every page and it being slanted anywhere else won't make sense.

PR #720, opened 2019-03-12
Enhancement/redirect to login

### LINKED ISSUE #693: App Should Show a 401 Page
When navigating to a Learning Object that is currently in the review stage for a collection that the user has access to, they are given a 404 page if their token has expired.

The app should instead indicate that they have been logged out and should sign in to view the requested content.

EVogel1999:
If a learning object exists and in working stage but the user is not logged in, it will navigate to the not-found page with a 401 status instead of a 404.  It will prompt users to login if they are not already logged in.

Closes #693 
Needs Cyber4All/learning-object-service#274

### DIFF
--- a/src/app/cube/details/details.component.ts
+++ b/src/app/cube/details/details.component.ts
@@ -146,8 +146,20 @@ export class DetailsComponent implements OnInit, OnDestroy {
        */
 
       if (e instanceof HttpErrorResponse) {
-        if (e.status === 500 || e.status === 401) {
+        if (e.status === 404) {
           this.router.navigate(['not-found'], { queryParams: { errorStatus: e.status } });
+        }
+        if (e.status === 401) {
+          let redirectUrl = '';
+          this.route.url.subscribe(segments => {
+            if (segments) {
+              segments.forEach(segment => {
+                redirectUrl = redirectUrl + '/' + segment.path;
+              });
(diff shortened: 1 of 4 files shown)


EVogel1999:
Screen Shot 2019-03-12 at 10 33 03 AM

EVogel1999:
What if we use the not-found component but refactor it to include all different types of error status codes?  This way we don't make multiple components that take up the same space if we choose to add more status codes to our backend (400 - 500).

EVogel1999:
Screen Shot 2019-03-14 at 10 44 40 AM
Screen Shot 2019-03-14 at 10 44 55 AM

src/app/cube/cube.routing.ts
```diff
@@ -65,6 +66,7 @@ const cube_routes: Routes = [
         data: { breadcrumb: 'Preferences' },
         canActivate: [AuthGuard]
       },
+      { path: 'details/error-status', component: ErrorStatusComponent },
```
DonSeannelly: @EVogel1999 what do you think about just having a template conditional to show this component? That way it doesn't route to a new page.

I believe this implementation was taken from the current 404 page, but it seems to be kind of confusing for users.

EVogel1999:
image

PR #746, opened 2019-03-19
Make clark-entity an internal package in the client with optional hasRevision flag

davidnso:
Before we can reference the hasRevision flag to switch between a working and released object, the clark-entity must be updated so it can properly reference this property. This PR... 

- adds clark-entity as an internal package in the service. 
- changes all imports that reference @cyber4all/clark-entity to reference the internal package
- Removes remnants of published and unpublished in the service. 
- adds the optional hasRevision flag to learning-objects in the client


src/types/rxjs.d.ts
```diff
@@ -0,0 +1,296 @@
+export class ArgumentOutOfRangeError {
```
nvisal1: Why was this rxjs file copied over?

package.json
```diff
@@ -54,8 +53,8 @@
     "@angular-devkit/build-angular": "~0.13.0",
     "@angular/language-service": "^7.2.3",
     "@types/file-saver": "^1.3.0",
-    "@types/node": "^10.5.2",
     "@types/jest": "^23.1.1",
+    "@types/node": "^10.5.2",
```
DonSeannelly: I see that this was here before, but I'm not sure we need this. The app isn't running inside node.

src/app/auth/register/gravatar-info/gravatar-info.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, OnInit, Input } from '@angular/core';
 import * as md5 from 'md5';
 import { FormGroup } from '@angular/forms';
-import { User } from '@cyber4all/clark-entity';
+import { User } from '../../../../entity/index';
```
DonSeannelly: This should be able to be 
```suggestion
import { User } from '../../../../entity';
```
because of the index file being there.

src/app/onion/dashboard/dashboard.component.ts
```diff
@@ -694,7 +694,6 @@ export class DashboardComponent implements OnInit, OnDestroy {
   async addToCollection(collection?: string) {
     if (collection) {
       this.collectionService.submit(this.focusedLearningObject.id, collection).then(() => {
-        this.focusedLearningObject.publish();
```
DonSeannelly: Can we verify that we don't have any template logic that uses this?

src/app/auth/register/gravatar-info/gravatar-info.component.ts
DonSeannelly: If we actually go with the type alias, this won't matter.

src/types/rxjs.d.ts
davidnso: Typesricpt was trying to fix an error where some imports were failing and I guess it created this file to add types. It'll be removed.

src/app/auth/register/gravatar-info/gravatar-info.component.ts
davidnso: Great idea on adding the path so we can use the alias ```@entity```. With that addition this shouldn't be necessary then. Would it be cleaner to reference entity as such ```"@entity": ["entity/index"]``` or still remove index

src/app/onion/dashboard/dashboard.component.ts
davidnso: from what I've seen from looking through and testing the dashboard after creating a learning object it doesn't look like there's any template logic there but @nrwinner might have more context.

jbarlow96: Did this get resolved?

DonSeannelly: It seems to be in order, but we might want to get @nrwinner to stage this on his machine and give it a run through to be sure.

davidnso: I agree

PR #755, opened 2019-03-25
Feature/remove bullets

### LINKED ISSUE #751: Strip Bullets from Outcomes
Many of our users are pasting their outcomes from bulleted lists inside of word documents. This results in outcomes where a bullet breaks up the sentence.
image

To increase readability, we should have the builder strip prefixed bullets from outcomes that are pasted in.

EVogel1999:
Remove bullets from outcomes so when a user pastes an outcome with a bullet the builder automatically strips it out of the outcome.

Closes #751

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "clark",
-  "version": "2.19.0",
+  "version": "2.20.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


EVogel1999:
Numbered bullets, small letter bullets with closing brace, usual bullet, upper case bullets with dots, and Roman numbered bullets.
Found the pattern on stack overflow here:
https://stackoverflow.com/questions/6505618/regular-expression-to-remove-bullets-from-the-text-is-not-working

PR #756, opened 2019-03-25
Wrap scaffold in scroll wrapper

### LINKED ISSUE #750: Add scrolling to children component in builder
If there is more children than what can fit on the display, the bottom of the list becomes unreachable because it renders off the screen. 
Screen Shot 2019-03-21 at 11 20 38 AM

EVogel1999:
Wrap scaffold component in a scroll wrapper, the same one used for the basic information section, so that if a user has many learning objects in the component then they can see all of the objects.

Closes #750


EVogel1999:
Before:
Screen Shot 2019-03-25 at 2 45 25 PM

After:
Screen Shot 2019-03-25 at 2 46 20 PM

src/app/onion/learning-object-builder/components/column-wrapper/column-wrapper.component.html
```diff
@@ -1,8 +1,10 @@
 <div class="builder-template-columns">
   <div #columnWrapper class="columns" [ngStyle]="{'top': columnOffset + 'px'}" [ngClass]="{'no-right': !rightColumn, 'no-left': !leftColumn}">
-    <div *ngIf="showColumn('l')" class="left-column column">
-      <div class="content">
-        <ng-content select="[left]"></ng-content>
+    <div class="scroll-wrapper" [ngStyle]="{'height': columnHeight + 'px'}">
```
pzalep1: How does this look on the outcomes and materials page? I think that the column is smaller on those pages, but I could be wrong.

EVogel1999: It's the same size but if the width is small the components looks shrinked, this is for both the info section and outcome.  Does not break the UI though.

Normal:
Screen Shot 2019-03-27 at 2 09 39 PM

Shrinked:
Screen Shot 2019-03-27 at 2 09 54 PM

PR #763, opened 2019-03-27
Redirect to dash from builder if editing released object

### LINKED ISSUE #616: Two requests being made for single object in builder
Resolve issue with builder firing off two requests for the same object

### LINKED ISSUE #750: Add scrolling to children component in builder

EVogel1999:
Redirects user to dashboard if they attempt to edit a released learning object by accessing the URL.  Instead of doing it in learning-object-builder guard which is what is making two requests for the same learning object.

Closes #616


src/app/onion/core/learning-object-builder.guard.ts
```diff
@@ -9,14 +9,6 @@ export class LearningObjectBuilderGuard implements CanActivate {
...
-
-    return this.learningObjectService.getLearningObject(next.paramMap.get('learningObjectId')).then(learningObject => {
-      if (learningObject.status && learningObject.status === LearningObject.Status.RELEASED) {
-        return false;
-      }
-      return true;
-    }).catch(error => {
-      return false;
-    })
+    return true;
```
nrwinner: If we're not using this Guard anymore can we delete it?

src/app/onion/dashboard/dashboard.component.ts
```diff
@@ -133,6 +135,13 @@ export class DashboardComponent implements OnInit, OnDestroy {
       const filters = {status: Array.from(this.filters.keys())};
       this.learningObjects = await this.getLearningObjects(filters);
     });
+
+    // if the user is trying to edit a released learning object, then get the status and alert the user
+    this.route.queryParams.subscribe(queryParams => {
```
nrwinner: This subscription should follow the takeUntil pattern we've used throughout the system to prevent leaving subscriptions open after components have been destroyed.

src/app/onion/learning-object-builder/learning-object-builder.component.ts
```diff
@@ -124,11 +124,15 @@ export class LearningObjectBuilderComponent implements OnInit, OnDestroy {
       // if name parameter found, instruct store to fetch full learning object
       if (id) {
         this.store.fetch(id).then(obj => {
-          // redirect user to dashboard if the object is in the working stage
-          if (!obj.status.includes('waiting') && !obj.status.includes('review') && !obj.status.includes('proofing')) {
-            this.setBuilderMode(obj);
+          if (obj.status === LearningObject.Status.RELEASED) {
```
nrwinner: Why not just add this check to the if statement below instead of an if-else here?

EVogel1999: If it is added to the if statement below, then the user will still be allowed to access editing via the builder if they are an editor/admin, I believe.

nrwinner: That's actually what we want, admins and editors do have privileged access to open other user's objects in the builder.

src/app/onion/core/learning-object-builder.guard.ts
EVogel1999: Guard is now deleted.

src/app/onion/dashboard/dashboard.component.ts
EVogel1999: Subscription now uses the takeUntil pattern.

Issue #759, opened 2019-03-27
Validate mailto link


DonSeannelly:
A unit test should be written to validate the existence of the mailto link, described in #644

Required by #644

Issue #643, opened 2019-02-19
Curators should be able to see users that are reviewers inside their collections


nrwinner:
As a curator, I want to be able to see which users are currently reviewers in each collection for which I'm a curator so that I can easily assign learning objects to them.

Issue #642, opened 2019-02-19
Curators should be able to add/remove reviewers inside their collections


nrwinner:
As a curator, I want to add/remove users as reviewers in collections for which I'm a curator so that I can control which users can review for each collection.

Required by #665

DonSeannelly:
Size 3

PR #766, opened 2019-03-28
Configure e2e tests in CI 

nvisal1:
This PR adds Circle CI configuration for e2e tests. 

Required by #759 
Required by #643 
Required by #642


.circleci/config.yml
```diff
@@ -12,25 +12,42 @@ jobs:
...
             - ./
   test:
     <<: *defaults
     steps:  
       - attach_workspace : 
           at: ~/clark-client
+      - run:
+          name: Clone e2eBox Repository 
+          command: >-
+           ls && cd .. && git clone https://${github_username}:${github_password}@github.com/nvisal1/CLARK-e2eBox.git
```
gshaw1997: Is this ls just for testing?

.circleci/config.yml
```diff
@@ -12,25 +12,42 @@ jobs:
...
+      - run:
+          name: Install Docker client
```
gshaw1997: Use `setup_remote_docker` instead.

https://circleci.com/docs/2.0/building-docker-images/

.circleci/config.yml
```diff
@@ -12,25 +12,42 @@ jobs:
...
+          command: |
+            set -x
+            VER="17.03.0-ce"
+            curl -L -o /tmp/docker-$VER.tgz https://download.docker.com/linux/static/stable/x86_64/docker-$VER.tgz
+            tar -xz -C /tmp -f /tmp/docker-$VER.tgz
+            mv /tmp/docker/* /usr/bin
+      - run: 
+          name: Install Docker Compose
```
gshaw1997: This will need to happen before setting up remote docker engine

https://circleci.com/docs/2.0/docker-compose/

cypress/integration/browse.spec.js
```diff
@@ -13,8 +13,9 @@ describe('Browse', () => {
     beforeEach(() => {
         // Return to home page before each test
         cy.fixture('route.json').then((route) => {
+            console.log(route)
             home = route;
```
gshaw1997: `home` should probably be renamed to be more indicative of the values it contains, seems to store more than just home info

src/environments/environment.ts
```diff
@@ -5,10 +5,10 @@
 
 export const environment = {
   production: false,
-  apiURL: 'http://localhost:3000',
+  apiURL: 'http://clark-gateway:3000',
```
gshaw1997: Are these temp values? How would this affect local dev?

DonSeannelly: This is going to force it to be run in a docker network. Can we get around this through any build commands with the Angular CLI?

If not, maybe we need to add some proxy setup to the image.

cypress/integration/browse.spec.js
```diff
@@ -13,8 +13,9 @@ describe('Browse', () => {
     beforeEach(() => {
         // Return to home page before each test
         cy.fixture('route.json').then((route) => {
+            console.log(route)
```
nrwinner: Why was this added?

PR #784, opened 2019-04-10
Implement client-side performance enhancements

nrwinner:
This PR introduces several measures to improve client performance:

1. Removes custom implementation of ngx-dropzone-wrapper and replaces it with the original version
2. Update code to be compatible with ahead-of-time (AOT) compiling instead of just-in-time (JIT)
3. Compress images and standard assets to reduce load time
4. Lazy load the Learning Object Builder to both improve performance and remove the arrow function syntax since it's incompatible with AOT.

The CI build command needs to be updated yet to run builds with AOT


DonSeannelly:
Can we update the CI config within this PR as well?

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           command: npm test -- auth/
       - run:
           name: Build Application
-          command: yarn ng build --prod --aot=false --build-optimizer=false --no-progress
+          command: yarn ng build --prod --no-progress
```
gshaw1997: :tada: Woohoo!

src/app/cube/home/components/splash/splash.component.ts
```diff
@@ -17,7 +17,7 @@ export class SplashComponent implements OnInit {
 
   ngOnInit() { }
 
-  submitSearch(event: KeyboardEvent) {
+  submitSearch(event: KeyboardEvent = undefined) {
```
gshaw1997: Comment please

nrwinner: No d2dedc8

PR #793, opened 2019-04-12
Change homepage splash image

nrwinner:
Screen Shot 2019-04-12 at 3 29 07 PM


nrwinner:
> So no blurred Blair?

I thought it looked better without the blur, but here's a blurred version for comparison.

Screen Shot 2019-04-12 at 3 45 25 PM

PR #801, opened 2019-04-19
Add View All back to Collections

DonSeannelly:
This PR adds the View All button back to the collection details page, as it was removed with the home page refactoring.


src/app/cube/collection-details/collection-details.component.ts
```diff
@@ -37,6 +41,7 @@ export class CollectionDetailsComponent implements OnInit, OnDestroy {
 
   async fetchCollection(name: string) {
     this.collection = await this.collectionService.getCollectionMetadata(name);
+    console.log(this.collection);
```
nrwinner: image

PR #802, opened 2019-04-19
Increase performance of the users page in collection dashboard

nrwinner:
This PR does a myriad of things targeted at increasing the performance of the collection dashboard.
1. General UI enhancements and clean up
2. Using `ChangeDetectionStrategy.OnPush` in "dumb" components https://blog.angular-university.io/onpush-change-detection-how-it-works/
3. Directly modifying the list of reviewers when one is added or removed instead of re-requesting the entire resource
4. Refactor the user cards to not render the menu list until the card is hovered and use javascript events instead of CSS events
5. Implement animations to improve overall aesthetic
6. Temporarily remove the broken learning objects page and the unfinished analytics page

https://i.imgur.com/NHUoNrR.mp4


src/app/admin/pages/users/users.component.html
```diff
@@ -29,12 +29,15 @@
 
   <div class="users" main>
     <div *ngIf="!loading; else loadingTemplate">
-      <div *ngIf="users?.length; else emptyTemplate" class="user-cards">
-        <clark-admin-user-card *ngFor="let user of users" class="user-cards__card"
+      <div *ngIf="users?.length; else emptyTemplate" class="user-cards" [@staggerChildren]>
+        <clark-admin-user-card *ngFor="let user of users;trackBy: trackby" class="user-cards__card"
```
nvisal1: Can you explain this change? Not sure what this does.

src/app/admin/pages/users/users.component.ts
```diff
@@ -33,9 +55,10 @@ export class UsersComponent implements OnInit {
     private route: ActivatedRoute,
     private toaster: ToasterService,
     public authService: AuthService,
-  ) { }
+  ) {
+  }
 
-  ngOnInit() {
+  ngAfterViewInit() {
```
DonSeannelly: What was the reason behind moving to this lifecycle hook?

nrwinner: I wanted the processor-intensive rendering to happen after the view was fully initialized instead of at the start of the process to avoid locking up the UI. It only really makes a difference in Chrome.

src/app/admin/pages/users/users.component.html
nrwinner: The`trackBy` function gives the `*ngFor` directive the ability to track changes to the list it represents. Without it, if I give `*ngFor` an array of learning objects and remove one of them, it doesn't know which one I removed so it re-renders the entire list. The `trackBy` function in the component Typescript takes an index and an objects and returns the property by which to track, in this case the learning object id. So now, if I add or remove an object from the list the directive knows what was changed and can re-render only that piece. More here https://netbasal.com/angular-2-improve-performance-with-trackby-cc147b5104e5

PR #806, opened 2019-04-24
Implement the collection dashboard

nrwinner:
This PR introduces a collection dashboard that provides administrators, editors, and curators a suite of functionality regarding users and learning objects in the CLARK system. There are too many screens to exhaustively cover with screenshots, I would ask that reviewers checkout the branch and play with the system.

Contributes to ch97

This is a big PR and I'm sorry :( if it makes you feel any better I too reviewed it so I can share in your pain.


nvisal1:
image

package.json
```diff
@@ -55,6 +53,7 @@
     "@types/node": "^10.5.2",
     "@types/uuid": "^3.4.3",
     "codelyzer": "^4.3.0",
+    "cypress": "^3.2.0",
```
pzalep1: Cypress should be removed before merging.

src/app/admin/components/filter-search/filter-search.component.ts
```diff
@@ -0,0 +1,267 @@
...
+                .width - 200
+          }
+        );
+      } else {
+        this.contextMenuService.destroy(this.filterMenu);
+      }
+
+      this.filterMenuDown = !!event;
+    } else {
+      console.error('Error! Attempted to use an unregistered context menu');
```
DonSeannelly: Can you elaborate on this error? I see it's a console error, so it's meant for the developer. What action should they take to remedy this?

src/app/admin/components/filter-search/filter-search.component.ts
```diff
@@ -0,0 +1,267 @@
...
+        return 'far fa-hourglass';
+      case 'review':
+        return 'far fa-sync';
+      case 'proofing':
+        return 'far fa-shield';
+      case 'released':
+        return 'far fa-eye';
+      case 'rejected':
+        return 'far fa-ban';
+    }
```
DonSeannelly: Could we pull this up to a higher level of the application where it could be shared across components/modules? It seems to be popping up a lot, so any changes will require updates in several places.

src/app/admin/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,52 @@
...
+   * Navigate to previous location
+   *
+   * @memberof SidebarComponent
+   */
+  navigateBack() {
+    this.router.navigateByUrl(this.history.lastRoute ? this.history.lastRoute.url : '/');
+  }
+
+  ngOnDestroy() {
+    // subscription clean up
```
DonSeannelly: I think this pattern is used enough throughout that we probably don't need a line comment 

```suggestion
```

src/app/admin/components/user-privileges/privileges-list/privileges-list.component.ts
```diff
@@ -0,0 +1,91 @@
+import { Component, OnInit, Input, Output, EventEmitter } from '@angular/core';
+import { first } from 'rxjs/operators';
+import { Subject } from 'rxjs';
+import { trigger, transition, style, animate, query, stagger } from '@angular/animations';
+
+@Component({
+  selector: 'clark-privileges-list',
+  templateUrl: './privileges-list.component.html',
+  styleUrls: ['./privileges-list.component.scss'],
+  animations: [
```
DonSeannelly: I think as animations grow we should identify a cutoff point where we place them in their own file. I don't think this one here is necessarily too big, but might want to be a guideline we establish

src/app/admin/components/user-privileges/privileges-list/privileges-list.component.ts
```diff
@@ -0,0 +1,91 @@
...
+   * Instructs the ngFor how to track changes to the list of privileges
+   *
+   * @param {*} _
+   * @param {*} item
+   * @returns
+   * @memberof PrivilegesListComponent
+   */
+  trackBy(_, item) {
+    return item.join('@');
+  }
```
DonSeannelly: I thought trackBy was a built in - can someone explain this to me?

src/app/admin/components/user-privileges/user-privileges.component.ts
```diff
@@ -0,0 +1,193 @@
...
+   *
+   * @private
+   * @memberof UserPrivilegesComponent
+   */
+  private getUserRoles() {
+    this.privilegeService.getRoles(this.user.id).then(roles => {
+      this.privileges = roles.map(x => x.split('@'));
+    this.getCollections();
+    }).catch(error => {
+      this.toaster.notify('Error!', 'There was an error fetching user\'s privileges. Please try again later.', 'bad', 'far fa-times');
```
DonSeannelly:
```suggestion
      this.toaster.notify('Error!', 'There was an error fetching this user\'s privileges. Please try again later.', 'bad', 'far fa-times');
```

src/app/admin/components/user-search-wrapper/user-search-wrapper.component.html
```diff
@@ -0,0 +1,34 @@
+<div class="reviewer-search__container">
+  <div class="container__user-dropdown">
+    <div class="user-dropdown__search">
+      <i class="fas fa-search search__icon"></i>
+      <input type="text" aria-label="user search" class="search__input" placeholder="Search Users" [(ngModel)]="query" (input)="userSearchInput$.next($event.currentTarget.value)">
```
pzalep1: aria-label should be aria-labelledby because the text is visible on the screen

src/app/admin/pages/learning-objects/learning-objects.component.ts
```diff
@@ -0,0 +1,274 @@
...
+    };
+    this.publicLearningObjectService
+      .getLearningObjects(query)
+      .then(val => {
+        this.learningObjects = val.learningObjects;
+      })
+      .catch(error => {
+        this.toaster.notify(
+          'Error!',
+          'There was an error fetching this user\'s learning objects. Please try again later.',
```
DonSeannelly:
```suggestion
          'There was an error fetching this user\'s Learning Objects. Please try again later.',
```

src/app/core/privilege.service.ts
```diff
@@ -0,0 +1,115 @@
...
+  private handleError(error: HttpErrorResponse) {
+    if (error.error instanceof ErrorEvent) {
+      // Client-side or network returned error
+      return throwError(error.error.message);
+    } else {
+      // API returned error
+      return throwError(error);
+    }
+  }
+}
```
DonSeannelly: Is this service being used outside of the collection dashboard? If not, we should consider putting it inside of a core module there to reduce the bundle size for the main app to load.

src/app/shared/navbar/navbar.component.ts
```diff
@@ -46,7 +47,7 @@ import {
   ]
 })
 export class NavbarComponent implements OnInit, AfterContentChecked, OnDestroy {
-  // FIXME: Convert 'class' to 'type' for consistancy
+  // FIXME: Convert 'class' to 'type' for consistency
```
DonSeannelly: What is the context of this FIXME? Is it saying to convert the `NavbarComponent` class to a type? I'm not really following.

src/environments/status-descriptions.ts
```diff
@@ -0,0 +1,43 @@
...
+    review(collection: string) {
+      return `This learning object is currently under review by the ${collection} review team. It is not yet released and cannot be edited until the review process is complete.`
+    },
+    proofing() {
+      return 'This learning object is currently undergoing proofing by the CLARK editorial team. It is not yet released and cannot be edited until the proofing process is complete.'
+    },
+    released(collection: string) {
+      return `This learning object is released in the ${collection} collection and can be browsed for.`;
+    }
+  };
```
DonSeannelly: This should be updated to `Learning Object`

src/environments/status-descriptions.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  constructor(private collectionService: CollectionService) { }
+
+  async getDescription(status: LearningObject.Status, collectionAbbreviation?: string) {
+    const needsCollection = [LearningObject.Status.RELEASED, LearningObject.Status.WAITING, LearningObject.Status.REVIEW];
```
DonSeannelly: I don't follow the `needsCollection` variable being a set of statuses? Am I missing some context, or is there a better name we could use here?

src/app/shared/dashboard-item/dashboard-item.component.html
```diff
@@ -1,10 +1,10 @@
 <div *ngIf="!hidden" class="dashboard-item" [ngClass]="{'disabled': disabled, 'shrink': small}">
   <div class="row-item">
-    <div>
+    <div *ngIf="hasCheckBox">
       <clark-checkbox *ngIf="!disabled" [setValue]="selected" (action)="toggleSelect($event)"></clark-checkbox>
```
pzalep1: Add aria-label to input

src/app/shared/popup-templates/collection-selector-popup/collection-selector-popup.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="banner">
+  <div class="banner__title">Select a Collection</div>
+  <div class="banner__sub">Please submit to CAE Community collection if not associated with other collections.</div>
+</div>
+<div class="collections-grid">
+  <clark-collections-grid (collections)="collections" [currentCollection]="currentCollection" (selected)="select($event)"></clark-collections-grid>
+  <div class="license-selection">
+    <input type="checkbox" name="license-agreement" aria-label="license agreement checkbox" [(ngModel)]="licenseAccepted"/>
```
pzalep1: aria-label="Clickable license agreement checkbox"

src/app/shared/popup-templates/collection-selector-popup/collection-selector-popup.component.scss
```diff
@@ -0,0 +1,67 @@
...
+    left: 0;
+    right: 0;
+    bottom: 0;
+    opacity: 0.04;
+    mask: $cube-pattern repeat;
+  }
+
+  .banner__title {
+    color: white;
+    font-size: 24px;
```
pzalep1: Font-size normalization. Larger would probably be the way to go.

src/app/admin/components/user-privileges/privileges-list/privileges-list.component.ts
nrwinner: The trackBy functionality is built in, but it takes a function to dictate its policy for tracking changes to items in the model. More information here: https://netbasal.com/angular-2-improve-performance-with-trackby-cc147b5104e5

src/environments/status-descriptions.ts
nrwinner: The `needsCollection` variable here dictates whether or not the status uses the name of a collection in its copy and 'needs" to fetch the full 'collection' from the CollectionService given its abbreviated name.

src/app/shared/navbar/navbar.component.ts
nrwinner: This FIXME was was there when I found it, I have no context. I just saw the typo and fixed it. It doesn't appear to apply to anything in the NavbarComponent (since that class cannot be converted to a type) so my guess is it can be removed.

src/app/shared/popup-templates/collection-selector-popup/collection-selector-popup.component.scss
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+.collections-grid {
+  margin-top: 100px;
+}
+
+.license-selection {
+  color: $dark-grey;
+  font-size: 18px;
```
pzalep1:
```suggestion
  font-size: $large;
```

src/app/admin/components/filter-search/filter-search.component.ts
nrwinner: The context menu component in its current state requires that any new context-menu's be registered with the context menu service before they can be used. Usually this occurs at the top of the file. I have a branch with this piece removed but I haven't finished it yet.

src/app/admin/pages/analytics/analytics.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
```
nvisal1: If this code isn't being used, it should be deleted.

src/app/admin/admin.component.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Retrieve a list of collection names for which the user is authorized to view (has curator status for)
+   *
+   * @memberof AdminComponent
+   */
+  async retrieveAuthorizedCollections() {
+    // wait for user to be logged in (edge case) and then fetch list of curated collections
+    return await this.authService.isLoggedIn.pipe(
+      skipWhile(x => x === false),
```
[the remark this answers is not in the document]
nrwinner: On slow connections, the app can load in an un authenticated state while it's waiting on the cookie to be validated with the service. When this happens, the accessGroups property is undefined and thus the user's authorized collections cannot be loaded.

src/app/admin/admin.component.ts
```diff
@@ -0,0 +1,121 @@
...
+      take(1)
+    )
+    .toPromise()
+    .then(async () => {
+      // we're sure the user is logged in here and so access groups should be defined
+      return await Promise.all(
+        this.authService.user['accessGroups']
```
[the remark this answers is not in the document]
nrwinner: In most cases, a User object in the client doesn't have an `accessGroups` property. This property only exists in the AuthService.

src/app/shared/collections-grid/collections-grid.component.html
```diff
@@ -1,29 +1,55 @@
...
       Loading collections...
     </div>
   </ng-template>
-  
+
   <ng-template #noneTemplate>
     <div class="none-template">
       <span><i class="far fa-times"></i></span>
-      Uh-oh! We weren't able to get a list of collections. Please try again later.
+      Uh-oh! We weren't able to get a list of collections. Please try again
```
nvisal1: It might be a good idea to reword this error message. "Uh-oh!" seems a little sarcastic.

package.json
```diff
@@ -8,9 +8,7 @@
     "start": "ng serve",
     "build": "ng build",
     "test": "jest --verbose  --coverage",
-    "lint": "ng lint",
-    "app": "cd .. cd ",
-    "e2e": "cypress open"
```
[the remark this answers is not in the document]
nvisal1: Cypress should not be installed as a dependency in the client anymore. The `e2e` script is not being used. The e2e box does not need this script for running tests in CI. It should be removed.

package.json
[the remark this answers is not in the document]
nvisal1: Both the cypress dependency and the command should be removed.

src/app/admin/admin.component.ts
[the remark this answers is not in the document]
DonSeannelly: Gotcha. Maybe in the future we could come up with a way to wait for the user to be validated before loading the components that are restricted when loading them directly.

src/app/admin/admin.component.ts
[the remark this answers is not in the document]
DonSeannelly: So in that case, since the `user` being read is from the `AuthService` where that property exists, shouldn't we have typings for that - allowing `this.authService.user.accessGroups`?

src/app/shared/collections-grid/collections-grid.component.html
nrwinner: @DonSeannelly since this is your copy, what are your thoughts? Personally I think I like it the way it is.

image

src/app/admin/components/filter-search/filter-search.component.ts
nrwinner: As far as I'm aware, this is the only place we're doing the icons like this. Perhaps some time should be taken to decide the best way to do the status icons across the platform, but as of right now I can't find any code that performs a similar function to this in the app.

src/app/admin/components/user-privileges/privileges-list/privileges-list.component.ts
nrwinner: I definitely agree with this.

src/app/admin/admin.component.ts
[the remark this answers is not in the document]
nrwinner: Oh actually we do already lol whoops

src/app/admin/admin.component.ts
```diff
@@ -0,0 +1,121 @@
...
+        .filter(group => group.includes('curator@'))
+        .map(group =>
+          this.collectionService.getCollection(group.split('@')[1]).then(c => this.authorizedCollections.push(c))
+        ))
+        .then(() => {
+          // remove the initialization block
+          this._initialized[0] = true;
+        });
```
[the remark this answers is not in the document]
DonSeannelly: I'm not sure if I'm fully following here. What I was thinking was that the anonymous function being used by the top level `then` (which responds to the promisified data from the pipe), is just one statement - a `Promise.all`. That statement though does quite a bit, and as an uniformed reader it's a little hard for me to grasp it.

My thought was that we could create a private helper method and pass that to the top level `then` that is used. So it would read:
```
.then(doTheThing)
```

This would allow for us to put JSDoc on that function to explain the purpose of the filter, mapping, and cleanup logic of the nested `then`.

src/app/shared/navbar/navbar.component.ts
DonSeannelly: I'm going to guess if we look at the git history for this file it was referring to something else that got removed and it didn't. My vote would be 🔥 it

src/environments/status-descriptions.ts
DonSeannelly: Ah that makes sense. Maybe we could add JSDoc to the function that says just that. It's pretty clear now, but I did not pick that up at first

DonSeannelly: Ah that makes sense. Maybe we could add JSDoc to the function that says just that. It's pretty clear now, but I did not pick that up at first

src/app/shared/collections-grid/collections-grid.component.html
DonSeannelly: I think we need to spend some time on identifying the tone and voice we want to use in copy and user-facing messages on the system. We alternate between fairly cold and robotic to somewhat playful at times.

If the "Uh-oh" seems a little off-putting maybe we could just drop it.

Also, question on this now that I'm looking at it again. Why did we remove `later .`?

src/app/admin/components/filter-search/filter-search.component.ts
DonSeannelly: I was thinking of `LearningObjectStatusIndicatorComponent` in my head, but it turns out I never went back to refactor the template so it's just making a ton of icon tags and ng-ifing them

src/app/shared/collections-grid/collections-grid.component.html
nrwinner: We didn't, later is just pushed down onto the next line for some reason

src/app/admin/admin.component.ts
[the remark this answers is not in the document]
nrwinner: At that point though, isn't the parent function just waiting for the user to be logged in and then handing off to another function? It's not handling any of the actual logic for the use case.

src/environments/status-descriptions.ts
nrwinner: Can do

src/app/admin/components/user-privileges/privileges-list/privileges-list.component.ts
nrwinner: I was noodling on this at lunch, I'm thinking it might be a good idea to decree that all animations should _always_ live in a separate file. A cutoff is a bit of a slippery slope I feel.

src/app/admin/admin.component.ts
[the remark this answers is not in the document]
DonSeannelly: Yes, that's correct. It would be setting things up and then invoking another function. Functions are a unit of decomposition that should be much smaller than a use case. We don't actually ever **need** a function, but they can significantly help in the way that our minds process code by reducing scope.

My suggestion here is just rooted in an opinion that promises in promises are not inherently readable - it reminds me of callback hell. I do think it would be best to decompose it, but since we don't have any guidelines on this as a team I can't say that I believe we must break it down.

I'll leave it up to your judgement.

src/app/admin/components/user-privileges/privileges-list/privileges-list.component.ts
DonSeannelly: I like it 👍

src/app/shared/collections-grid/collections-grid.component.html
nrwinner: Oh, and we absolutely need a voice-and-tone guideline set. Will be part of the forever impending style guide I keep promising.

DonSeannelly: @nrwinner maybe we could break it out of that so the can doesn't get kicked too far down the road. It seems to be something that is coming up on a weekly basis. I'd like to get even just a preliminary draft put together in the next few weeks.

nrwinner: I've removed the 'Uh-oh' for the time being until we solidify our voice and tone guidelines

src/app/admin/components/user-privileges/privileges-list/privileges-list.component.ts
nrwinner: Collection-dashboard animations are extracted to their own files

package.json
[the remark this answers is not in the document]
nrwinner: Cypress is definitively not still in the package.json as a dependency, this must be an outdated file view.

PR #807, opened 2019-04-25
Direct Upload to S3 using AWS S3 SDK

gshaw1997:
This PR removes dropzone and replaces file uploads via CLARK service with direct uploads to S3 using the client AWS S3 SDK.

**Notable changes**
- Complete Removal of Dropzone
- Usage of drag and drop / File API for processing folder and file uploads
- Uploads using S3 SDK 
- AWS Resource credential management using Cognito
- Token payload update. See https://github.com/Cyber4All/user-service/pull/137
- Simplified upload progress indicator.

**UI Changes**
- Upload progress indicator
Screenshot from 2019-04-24 20-05-21
Screenshot from 2019-04-24 20-06-00

**WIP**
- Implement sliding window for processing uploads from queue to avoid browser locking. :heavy_check_mark: 
- Potentially implement user triggered upload retries? @Cyber4All/developers  I would like thoughts on implementing this feature? Do we want to go this route/ should this feature go into this PR? :question: 

**Update WIP**
- Implement queue based processing of requests to add file meta. Sending 1000+ `FileMeta` in a single request causes service error out and not save metadata. :heavy_check_mark:


src/app/onion/learning-object-builder/components/content-upload/app/services/file-management.service.ts
```diff
@@ -0,0 +1,234 @@
...
+        IdentityId,
+        Logins
+      },
+      { region: environment.cognitoRegion }
+    );
+    this.S3 = new AWS.S3({ region: environment.s3BucketRegion });
+  }
+
+  /**
+   * Sends learning object ID and file name to API for deletion.
```
DonSeannelly: I'm assuming that the service handles the file deletion still. Maybe now that we have a mix of client and server accessing S3, the JSDoc could specify that this is what happens, and so we don't need to manually connect to S3 for this.

src/app/onion/learning-object-builder/components/content-upload/app/upload/upload.component.html
```diff
@@ -2,54 +2,97 @@
...
+        class="slide slide1"
+        (dragenter)="dragenter($event)"
+        (dragover)="dragover($event)"
+        (drop)="handleDrop($event)"
+      >
         <div *ngIf="solutionUpload" class="solutions-warning">
-          Note: While the upload of solution files are not prohibited, they won't be included during a learning
-          object's download.
+          Note: While the upload of solution files are not prohibited, they
+          won't be included during a learning object's download.
```
DonSeannelly: For consistency, we're trying to make sure that "Learning Object" is capitalized throughout the UI:

```suggestion
          won't be included during a Learning Object's download.
```

src/app/core/auth.service.ts
```diff
@@ -22,10 +22,21 @@ export enum AUTH_GROUP {
   ADMIN
 }
 
+export interface Tokens {
```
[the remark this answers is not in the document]
gshaw1997: Yes, this is the login response. Wasn't set on naming, but went with `Tokens` because it matched the resource being created on login and registration and being requested during refresh.

src/app/onion/learning-object-builder/components/content-upload/app/services/file-management.service.ts
```diff
@@ -0,0 +1,234 @@
...
+          }
+          remainingUploads--;
+          if (remainingUploads === 0) {
+            uploadUpdate$.next({
+              type: 'queueComplete',
+              data: { successful, failed }
+            });
+          }
+        }
+      ).on('httpUploadProgress', evt => {
```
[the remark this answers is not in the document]
gshaw1997: Errors are tracked in the callback. The only listener the API exposes is this `httpUploadProgress` event.

gshaw1997:
> A few changes and questions noted here.
> 
> **In reference to the WIP comments:**
> Does the browser just lock, or fail? If it is only lockup, I'd like to implement the window in a new PR, to reduce load on reviewers.
> 
> I think the retries are a nice to have, since we don't implement that now. Let's log a feature request and figure out where that will fit in our priorities.

**Browser Lock**

gshaw1997:
Requesting a review from all @Cyber4All/developers to test uploads when you all get a chance. Thank you!

src/app/onion/learning-object-builder/components/content-upload/app/upload/upload.component.ts
```diff
@@ -455,24 +455,32 @@ export class UploadComponent implements OnInit, AfterViewInit, OnDestroy {
...
-      .subscribe(update => this.handleUploadUpdates(update));
+    try {
+      this.fileManager
+        .upload(this.bucketUploadPath, files)
+        .subscribe(update => this.handleUploadUpdates(update));
+    } catch (e) {
+      if (e.name === UploadErrorReason.Credentials) {
+        this.handleCredentialsError();
+      } else {
+        this.error$.next('Unable to upload files at this time.');
```
DonSeannelly: Is this a user-facing error? If so, we should clearly have a subject in the sentence to make sure that it is clear **we** are unable to upload files and the user should try again later.

src/app/onion/learning-object-builder/components/content-upload/app/upload/upload.component.ts
```diff
@@ -594,19 +602,21 @@ export class UploadComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    if (update.error.name === UploadErrorReason.Credentials) {
+      this.handleCredentialsError();
     } else {
       const index = this.uploadQueueMap[update.data.fullPath];
       this.uploadQueue[index].success = false;
     }
   }
 
+  private handleCredentialsError() {
+    this.error$.next(`Cannot upload files at this time. Invalid credentials.`);
```
DonSeannelly: What action should the user take at this point?

src/app/onion/learning-object-builder/components/content-upload/app/upload/upload.component.ts
```diff
@@ -594,19 +602,21 @@ export class UploadComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    const s = this.uploadQueue.filter(file => !file.success);
+    // TODO: Fetch new credentials and attempt retry?
```
DonSeannelly: Is there a plan to implement this TODO?

src/app/core/auth.service.ts
```diff
@@ -49,19 +55,71 @@ export class AuthService {
...
-      this.validate().then(
-        _ => {
-          this.changeStatus(true);
-        },
-        _ => {
-          this.cookies.remove('presence');
-          this.changeStatus(false);
-        }
-      );
+      this.validate();
```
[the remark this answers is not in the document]
gshaw1997: All logic is handled inside of the `validate` function now instead of the class's constructor

src/app/core/auth.service.ts
```diff
@@ -49,19 +55,71 @@ export class AuthService {
...
+
+  /**
+   * Unset session related data
+   *
+   * @private
+   * @memberof AuthService
+   */
+  private endSession() {
+    this.user = undefined;
+    this.openIdToken = null;
```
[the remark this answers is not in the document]
gshaw1997: user being set to undefined was old code that was copied and not updated. I'll update it here.

src/app/core/auth.service.ts
```diff
@@ -166,18 +224,19 @@ export class AuthService {
...
           catchError(this.handleError)
         )
         .toPromise();
-      this.user = response.user as AuthUser;
-      this.openIdToken = response.openId;
+      this.user = response;
+      this.changeStatus(true);
       this.assignUserToGroup();
     } catch (error) {
+      this.endSession();
```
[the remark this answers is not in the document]
gshaw1997: Yes that is correct

src/app/onion/learning-object-builder/components/content-upload/app/upload/file-upload-status/file-upload-status.component.scss
```diff
@@ -1,29 +1,92 @@
...
+
+clark-progress {
+  z-index: 2;
+  height: auto;
+  &.completed {
+    height: 0;
+    opacity: 0;
+  }
+}
+:host::ng-deep clark-progress {
```
nrwinner: Is there any better way to do this? `::ng-deep` scares me a little bit

src/app/onion/learning-object-builder/components/content-upload/app/upload/upload.component.html
```diff
@@ -2,54 +2,98 @@
...
       <div class="uploader-content">
         <div class="uploader-icon">
-          <i class='far fa-arrow-up'></i>
+          <i class="far fa-arrow-up"></i>
         </div>
         <div class="uploader-text">
           <div>Drop files and folders here</div>
-          to add them to {{ (!openPath) ? 'your learning object' : '\'' + openPath + '\''}}!
+          to add them to
+          {{ !openPath ? 'your learning object' : "'" + openPath + "'" }}!
```
nrwinner: Awesome!

nrwinner:
> The code here looks good. After playing around with this branch locally, I did notice a few things:
> 
> 1. If a network error occurs during upload, the modal kicks the user back to the dashboard even if the close the modal instead of clicking the button. Does this workflow make sense? Or would it be better to let them stay on the current page and fix their connection?
> 2. Very odd issue here. If I click the upload button, and select to upload a file, I get the system dialog to pop up - this is normal. However, if I double click a file in the dialog it breaks all file API functionality and kicks my CPU into high gear. The UI is still responsive for the application, but clicking the list items from the upload button and drag/drop to the browser do nothing (this is on MacOS)

I can't reproduce this @DonSeannelly

PR #813, opened 2019-05-14
Hot fix/validate urls

pzalep1:
This PR is to address the issue with URLs either being empty or do not contain https:// at the beginning. In both of these cases when they show up on the details page they are broken links. 

Case 1: A new url row has been added and both title and url are empty. If the user leaves the page then the last one does not save, preventing an empty url field. The add new button is also disabled so a new empty row cannot be added. This prevents an empty row being added in between two valid rows. 
Screen Shot 2019-05-14 at 12 10 50 PM

Case 2: A new url row has been added and the title has been populated. If the user leaves that row then an error message occurs. 
Screen Shot 2019-05-14 at 11 25 29 AM
Case 3: A new url row has been added and the title has been populated and the url field is not valid. 
Screen Shot 2019-05-14 at 11 23 54 AM
Case 4: A new url row is added and the title and url field have been populated. The index saves and a saved message appears. 

** Highlight new row so that the new row can be more easily identified in long lists of URLs. The highlight will disappear once both fields have been touched. 
Screen Shot 2019-05-20 at 2 15 22 PM
Screen Shot 2019-05-20 at 2 15 40 PM


pzalep1:
@DonSeannelly I fixed this it now accepts http:// and will save the url.

PR #817, opened 2019-05-17
Feature/add guidelines filter

pzalep1:
This PR adds the guideline sources to the filter panel to filter on the browse page.
Screen Shot 2019-05-21 at 11 03 16 AM

** The CAE CDE 2019 guidelines no longer show up. They are filtered out in this PR


src/app/cube/browse/browse.component.ts
```diff
@@ -428,6 +438,7 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
   }
 
   async fetchLearningObjects(query: Query) {
+    console.log(query);
```
DonSeannelly: Looks like a dev log was left here
```suggestion
```

PR #827, opened 2019-05-31
Optimize DirectoryTree

gshaw1997:
This PR refactors and optimizes logic inside of the `DirectoryTree` to reduce the time complexity of most operations to O(1), and the traversal of the tree to O(n). This was accomplished through the use of maps to cache node and file locations which reduced iteration and unnecessary recursion.

Additionally, this PR optimizes file browser navigation by only traversing the tree for a node if the user jumps more that one directory which only occurs when the user navigates via breadcrumbs.
If the user is navigating folder to folder, the file browser uses the current node's cache map of locations to get the next node in constant time. Similarly, if the user navigates via breadcrumbs and only jumps one directory, the parent of the current node is returned in constant time.

**Note** The diff for the `DirectoryTree` is large, so GitHub collapsed the view by default, but most changes have been made here https://github.com/Cyber4All/clark-client/compare/refactor/file-directory?expand=1#diff-6dd9d9601f9431816929bc1fde3e0856

### DIFF
--- a/src/app/shared/filesystem/DirectoryTree.ts
+++ b/src/app/shared/filesystem/DirectoryTree.ts
@@ -8,15 +8,23 @@ import { LearningObject } from '@entity';
  * @class DirectoryTree
  */
 export class DirectoryTree {
-  private _root: DirectoryNode;
-  private lastNode: DirectoryNode;
-  private fileMap: Map<string, { path: string; file: string }> = new Map<
-    string,
-    { path: string; file: string }
-  >();
-  private pathMap: Map<string, number> = new Map<string, number>();
+  private root: DirectoryNode;
+  private lastTouchedNode: DirectoryNode;
+
   constructor() {
-    this._root = new DirectoryNode('', '', null);
+    this.root = new DirectoryNode('', '', null);
+  }
+
+  /**
+   * Returns folder DirectoryNode object if the folder exists within parent DirectoryNode
+   *
+   * @param {DirectoryNode} parent [The folder that contains the sub folder to be returned]
+   * @param {string} folderName [The name of the sub folder that should be returned]
+   * @returns {DirectoryNode}
+   * @memberof DirectoryTree
+   */
+  public getFolder(parent: DirectoryNode, folderName: string): DirectoryNode {
+    return parent.getFolder(folderName);
   }
   /**
    * Adds new Files to Directory Tree
@@ -25,23 +33,9 @@ export class DirectoryTree {
    * @memberof DirectoryTree
    */
   public addFiles(files: LearningObject.Material.File[]) {
-    // Returns new files and files that have had their metadata changed
-    const getUpdatedFiles = (fileList: LearningObject.Material.File[]) =>
-      fileList.filter(file => {
-        const mappedFile = this.fileMap.get(file.id);
-        if (mappedFile && mappedFile.file === JSON.stringify(file)) {
-          return false;
-        }
-        return true;
-      });
-    const newFiles = getUpdatedFiles(files);
-    for (const file of newFiles) {
-      this.fileMap.set(file.id, {
-        path: file.fullPath ? file.fullPath : file.name,
-        file: JSON.stringify(file)
-      });
+    files.forEach(file => {
       if (!file.fullPath) {
-        this._root.addFile(file);
+        this.root.addFile(file);
       } else {
         const paths = getPaths(file.fullPath);
         let node = this.traversePath(paths);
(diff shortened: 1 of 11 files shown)


src/app/shared/filesystem/DirectoryTree.ts
```diff
@@ -382,17 +314,29 @@ export class DirectoryNode {
...
+  public removeFolder(folderName: string): DirectoryNode {
+    const index = this.folderMap.get(folderName);
+    if (index >= 0) {
+      const deleted = this.folders[index];
+      this.folders.splice(index, 1);
+      this.folderMap.delete(folderName);
+      this.reIndexCacheMap(index, this.folders, this.folderMap);
+      return deleted;
+    }
+    return null;
```
DonSeannelly: Question on the return `null`'s: Why did you choose to explicitly return null, compared to just letting undefined come back as default?

src/app/shared/filesystem/file-browser/file-browser.component.html
```diff
@@ -1,8 +1,8 @@
 <div class="top-wrapper">
-  <clark-file-breadcrumb [paths$]="currentPath$"></clark-file-breadcrumb>
+  <clark-file-breadcrumb [paths]="currentPath" (pathChanged)="handlePathChanged($event)"></clark-file-breadcrumb>
   <div class="view-switcher inline-flex right">
     <div class="inline-flex">
-      <button *ngIf="canManage && ((currentNode$| async)?.children.length || (currentNode$| async)?.files.length)" id="new-file" (click)="emitNewOptionClick($event)" class="button new">
+      <button *ngIf="canManage && ((currentNode$| async)?.folders.length || (currentNode$| async)?.files.length)" id="new-file" (click)="emitNewOptionClick($event)" class="button new">
         New Upload <i class="fas fa-upload"></i>
       </button>
       <!-- <span *ngIf="view==='list'" (click)="view='grid'" class="button clean">
```
DonSeannelly: Does this component still have the grid view, we just don't allow it to be turned on?

src/app/shared/filesystem/file-browser/file-browser.component.ts
```diff
@@ -122,22 +116,41 @@ export class FileBrowserComponent implements OnInit {
...
+    this.emitCurrentNode(node);
   }
+
   /**
-   * Open node at current path
+   * Determines what node to refresh the view with by checking the path change
+   * If the path decreased in length by 1, we know that the user clicked one directory higher from breadcrumbs
+   * So, no traversal is needed. Just emit the parent node
+   *
+   * If the path changed by more than 1, the user jumped multiple directories and we need to traverse to find node to emit
```
DonSeannelly: These docs are great 👍

src/app/shared/filesystem/DirectoryTree.ts
gshaw1997: @DonSeannelly Since the return type is defined all execution branches must return.

src/app/shared/filesystem/file-browser/file-browser.component.html
gshaw1997: Good catch! We don't have the grid view anymore. I'll remove the commented code here.

src/app/shared/filesystem/DirectoryTree.ts
```diff
@@ -52,75 +46,33 @@ export class DirectoryTree {
...
     let lastCreatedNode: DirectoryNode;
     for (let i = continueFromIndex; i < paths.length; i++) {
-      const p = paths[i];
-      _paths.push(p);
-      const _node = this.traversePath(_paths);
-      if (!_node) {
-        const name = p;
-        lastCreatedNode = this.add(name, `${_paths.join('/')}`, this.lastNode);
-      }
+      const folderName = paths[i];
```
nrwinner: Bless you for this variable rename

src/app/shared/filesystem/file-browser/file-browser.component.html
```diff
@@ -1,28 +1,22 @@
...
-      <span *ngIf="view==='grid'" (click)="view='list'" class="button clean">
-        <i class="fas fa-list icon-clean"></i>
-      </span> -->
     </div>
   </div>
 
 </div>
-<div (click)="emitContainerClick($event)" class="flex-wrapper" [ngClass]="{empty: !(currentNode$| async)?.children.length && !(currentNode$| async)?.files.length}">
-  <div *ngIf="canManage && !(currentNode$| async)?.children.length && !(currentNode$| async)?.files.length" class="dropzone-message">
+<div (click)="emitContainerClick($event)" class="flex-wrapper" [ngClass]="{empty: !(currentNode$| async)?.folders.length && !(currentNode$| async)?.files.length}">
```
nrwinner: Why not use the new `isFolderEmpty` function defined above here for the `empty` class?
```suggestion
<div (click)="emitContainerClick($event)" class="flex-wrapper" [ngClass]="{empty: isFolderEmpty(currentNode$| async)}">
```

src/app/shared/filesystem/file-browser/file-browser.component.html
```diff
@@ -1,28 +1,22 @@
...
+  <div *ngIf="canManage && !(currentNode$| async)?.folders.length && !(currentNode$| async)?.files.length" class="dropzone-message">
```
nrwinner: Why not use the new `isFolderEmpty` function here?
```suggestion
  <div *ngIf="canManage && isFolderEmpty(currentNode$| async)" class="dropzone-message">
```

src/app/shared/filesystem/file-browser/file-browser.component.html
```diff
@@ -1,28 +1,22 @@
...
     <i class="fal fa-cloud-upload"></i>
     <div class="empty-title">No materials here!</div>
     <button class="button" id="new-file" (click)="emitNewOptionClick($event)">Click here to upload</button>
     <span>or drag and drop files and folders here</span>
   </div>
-  <clark-file-list-view *ngIf="((currentNode$| async)?.children.length || (currentNode$| async)?.files.length)"
+  <clark-file-list-view *ngIf="((currentNode$| async)?.folders.length || (currentNode$| async)?.files.length)"
```
nrwinner: Why not use the new isFolderEmpty function here?
```suggestion
  <clark-file-list-view *ngIf="!isFolderEmpty(currentNode$| async)"
```

PR #840, opened 2019-06-11
Update collection page route

nvisal1:
This PR updates the route for the collection page.

Collections are stored with abbreviated names (abvName). These are used as unique identifiers for collections. 

The current route structure for the collection page is
https://clark.center/c/NSA%20NCCP

Instead of using the full name, we should use the abbreviated name. 

The same route is now structured as
https://clark.center/c/nccp

@nrwinner @pzalep1 
Did I miss any links to the collection page? I updated the links on the home and details pages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,11 +2,16 @@
   "name": "clark",
   "version": "2.32.2",
   "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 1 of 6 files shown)


pzalep1:
There is also one used in the details-splash-component

PR #841, opened 2019-06-12
Organization by User Pie Chart

jbarlow96:
This PR adds a pie chart that distributes the quantity of users by organization. It also adds tab indexing to assist with 508 Compliant requirements.


pzalep1:
Screen Shot 2019-06-13 at 9 30 07 AM

jbarlow96:
Screen Shot 2019-06-13 at 10 02 37 AM

src/app/cube/usage-stats/usage-stats.component.ts
```diff
@@ -121,6 +125,54 @@ export class UsageStatsComponent implements OnInit {
       ];
   }
 
+  /**
+   * Constructs chart for Organization Breakdown distribution
+   *
```
DonSeannelly: We should add a fix me to the docs here so that it’s clear that this data is hardcoded and should eventually be pulled from our backend so that it can update dynamically

PR #848, opened 2019-06-18
Fix bug where directory tree is rebuilt and current path reset to root when files are added

nrwinner:


src/app/onion/learning-object-builder/components/content-upload/app/file-manager/file-manager.component.ts
```diff
@@ -149,17 +164,42 @@ export class FileManagerComponent implements OnInit, OnDestroy {
...
+
+        if (isFolder) {
+          node.removeFolder(name);
+        } else {
+          node.removeFile(name);
+        }
+
+        this.currentNode$.next(node);
+      } else {
+        console.log('ya messed up')
```
gshaw1997: https://img.memecdn.com/come-on-man_o_1094311.jpg

PR #850, opened 2019-06-20
Revisions/add builder banner

pzalep1:
The purpose of this PR is to add a banner to the builder to notify the user that they are making a revision to a released object. 

Unreleased Revised Version:
Screen Shot 2019-06-20 at 10 38 59 AM

Unreleased Unrevised Version: 
Screen Shot 2019-06-20 at 11 23 58 AM


src/app/onion/learning-object-builder/components/builder-navbar/builder-navbar.component.ts
```diff
@@ -31,6 +31,8 @@ export class BuilderNavbarComponent implements OnDestroy {
   learningObject: LearningObject;
   collection: Collection;
   statusDescription: string;
+  // FIXME: This will need to set based on the data coming back once the service is in place
+  revisedVersion = false;
```
DonSeannelly: How do we plan to determine this? Will this component have a function that evaluates this? I think it could be a good idea to go as far as we can at the moment to encapsulate the fixme to a smaller scope.

pzalep1: @DonSeannelly I am not quite sure what we are going to use to determine this. I can write a function to evaluate this, but I am not sure what is going to be used to determine if the object being opened is an original or revision.

pzalep1:
> Does this animate correctly in with the builder's navbar?

No, this does not. I will make the change

PR #858, opened 2019-06-27
Configure Lighthouse in CI

nvisal1:
***Purpose of this PR***
Add job to Circle CI that runs Lighthouse tests

***General Changes***
 - The Client image is built and published to Dockerhub in a new `publish-image` job.
Lighthouse tests are run after that in the `lighthouse` job.
 - A Slack message is sent after test completion. 
 - This PR also removes Cypress

***Circle CI Workflow Changes***
build-and-test:

The unit test step was removed from the `build` job and placed into the `test` job. If the app builds, then Circle CI publishes a new image tag to Dockerhub.
Screen Shot 2019-07-05 at 1 32 30 PM

***Slack Message***
Screen Shot 2019-07-03 at 3 10 01 PM


nvisal1:
> Do we have a threshold set on audit results that will fail builds if the results of the audits are below expected?

I was planning to these as the threshold values
image

nrwinner:
> > Do we have a threshold set on audit results that will fail builds if the results of the audits are below expected?
> 
> I was planning to these as the threshold values
> image

We need to be careful with that, I believe the Lighthouse CLI defaults to simulated network throttling, while those were taken using applied network throttling. There is a massive difference between the results of the different throttling for some unknown reason.

nrwinner:
As for audits I'd like to see, I'm most worried about Performance and Accessibility categories. It's important to note, however, that the accessibility category is far from exhaustive as Google admits. It does not negate the need for manual accessibility reviews of code.

DonSeannelly:
> > > Do we have a threshold set on audit results that will fail builds if the results of the audits are below expected?
> > 
> > 
> > I was planning to these as the threshold values
> > image
> 
> @nvisal1 @gshaw1997 We need to be careful with that, I believe the Lighthouse CLI defaults to simulated network throttling, while those were taken using applied network throttling. There is a massive difference between the results of the different throttling for some unknown reason.

@nrwinner I started looking into this to help @nvisal1 with some of the work on his plate. I think this link is rather informative on the subject: https://github.com/GoogleChrome/lighthouse/blob/master/docs/throttling.md

Additionally, I did some reading here: https://blog.checklyhq.com/how-we-got-a-100-lighthouse-performance-score-for-our-vue-js-app/ and I wonder what our aims are with throttling, seeing that half the app is not even meant for mobile usage.

Maybe before we pick a means of throttling in our testing, we get some data on mobile usage of the system and what we want to set as a baseline?

PR #871, opened 2019-07-05
update view file message

jbarlow96:
**_Purpose of this PR_**
To change the message that appears when attempting to view a file while not being signed in to make it more clear that you have to be signed in to be view the file. It also let's the user (whether signed out or signed in) know if the file they are trying to preview can't be previewed.

**Message Displayed When Signed In and the File Type can't be Previewed**
image

**Message Displayed While Being Signed Out and the File Type can be Previewed**
image

**Messaged Displayed While Being Signed Out and the File Type can't be Previewed**
image


src/app/shared/filesystem/file-preview/file-preview.component.html
```diff
@@ -1,4 +1,4 @@
 <h3 tabindex="0">
-  <i class="fas fa-exclamation-circle"></i> Preview is not available at this time. Click the "Download Now" button located above to download this object.
+  <i class="fas fa-exclamation-circle"></i> Please log in to preview file. Click the "Download Now" button located above to download this object.
```
DonSeannelly: I think that this will also show if a user is logged in and clicks on something like a `.vmdk` file of which we don't have a preview. We should have this check if the user is logged in or not and display a message accordingly.

src/app/shared/filesystem/file-preview/file-preview.component.ts
```diff
@@ -1,15 +1,23 @@
...
+import { Subject } from 'rxjs';
+import { takeUntil } from 'rxjs/operators';
 
 @Component({
   selector: 'clark-file-preview',
   templateUrl: './file-preview.component.html',
   styleUrls: ['./file-preview.component.scss']
 })
 export class FilePreviewComponent implements OnInit {
+  private isDestroyed$ = new Subject<void>();
```
DonSeannelly: Looks like the `isDestroyed` pattern has started being used here. We should be sure to add ngOnDestroy in the following way to clean everything up:

```
ngOnDestroy() {
  this.isDestroyed$.next();
  this.isDestroyed$.unsubscribe();
}
```

nrwinner:
@jbarlow96 The verbiage for the signed out use-case seems a bit confusing. "Please log in...", "Click the download button above...", are we instructing the user in a two-step process? If they're not signed in, there is no download button above.

PR #874, opened 2019-07-08
508 Navigation Bar

pzalep1:
The purpose of this PR is to make changes to the navbar in order for it to be accessible. There were no visual changes although some divs were changed to the semantic HTML elements that they were imitating.


src/app/shared/contextmenu/context-menu/context-menu.component.ts
```diff
@@ -63,6 +63,12 @@ export class ContextMenuComponent implements AfterViewInit, OnDestroy {
     const domElem = (this.viewer.hostView as EmbeddedViewRef<any>)
       .rootNodes[0] as HTMLElement;
 
+    // for each
```
DonSeannelly: This comment seems a little vague. Perhaps we could elaborate or remove it.

src/app/shared/contextmenu/context-menu/context-menu.component.ts
```diff
@@ -105,6 +111,28 @@ export class ContextMenuComponent implements AfterViewInit, OnDestroy {
...
+      domElem.querySelectorAll('li:first-child, li:last-child').forEach(el => {
+
+        el.addEventListener('blur', (event: FocusEvent) => {
+
+          if (!event.relatedTarget || (event.relatedTarget as HTMLElement).parentElement !== firstElement.parentElement) {
+            this.close.emit();
+          }
+        });
+      });
+
```
DonSeannelly: @nrwinner can we add a doc comment here to describe this pattern a little more thoroughly?

src/app/cube/home/components/splash/splash.component.html
```diff
@@ -1,12 +1,12 @@
...
   <div class="splash__search">
-    <div class="search__text subtitle-text">
+    <div class="search__text subtitle-text" tabindex="0" id="pageContent">
       Effective cybersecurity curriculum at your fingertips.
     </div>
     <div class="search__input">
       <i class="far fa-search"></i>
       <input
-        aria-label="Search clark"
+        id="pageContent"
```
nrwinner: What's happening here? Why does the input have the same id as the div above it? Additionally, why remove the aria-label?

src/app/shared/mappings-filter/mappings-filter.component.html
```diff
@@ -6,13 +6,13 @@
       </div>
     </ng-container>
     <input #searchInput [ngClass]="{'transparent': focusedOutcome && !focused}" type="search" placeholder="Search for mappings..." aria-label="Search for Mappings" (focus)="focusInput(searchInput.value)" (blur)="focused = false;" />
-    <div class="search-sources" id="search-sources" [ngClass]="{'active': source}" tip="Filter results by sources" tipLocation="bottom" (click)="sourcesDown = true; preventClose('sources')">
+    <button class="search-sources" id="search-sources" [ngClass]="{'active': source}" aria-label="Filter results by sources" tip="Filter results by sources" tipLocation="bottom" (click)="sourcesDown = true; preventClose('sources')" (keypress)="sourcesDown = true; preventClose('sources')">
```
nrwinner: Can we remove the keypress directive here? That event will fire for ANY key that gets pressed, not just one of the activation keys (Enter or Space). The new activate directive should make this superfluous anyway 😄

src/app/shared/navbar/navbar.component.html
```diff
@@ -19,36 +22,36 @@
...
             <i class="fas fa-chevron-down"></i>
-          </div>
+          </button>
 
           <clark-context-menu *ngIf="contributorDown" [anchor]="contributorDropdownButtonElement" [offset]="{ top: 10, left: 5 }" (close)="contributorDropdown()">
             <div #contextMenu>
               <ul>
-                <li routerLink="/onion/dashboard" (click)="contributorDropdown()"><i class="far fa-list"></i>Your Dashboard</li>
-                <li routerLink="/onion/learning-object-builder" (click)="contributorDropdown()"><i class="far fa-plus"></i>Create a Learning Object</li>
+                <li tabindex="0" routerLink="/onion/dashboard" (click)="contributorDropdown()" id="contributorDropdown1"><i class="far fa-list"></i>Your Dashboard</li>
```
nrwinner: Could we potentially get a better ID here?
image

src/app/shared/navbar/navbar.component.scss
```diff
@@ -25,13 +40,15 @@
       position: absolute;
       bottom: -50px;
       color: white;
-      left: 0;
-      right: 0;
+      left: 50%;
```
nrwinner: image

src/app/cube/home/components/splash/splash.component.html
pzalep1: There is an aria-label attached to that input. I think it's on a different line and I don't know why that I did that.

src/app/shared/navbar/navbar.component.html
pzalep1: That has to go actually. Remnant of something I tried for the search

src/app/cube/home/components/splash/splash.component.html
pzalep1: There is an aria-label there a little ways down the block.

PR #876, opened 2019-07-09
Statistics Page Bootstrap adjustments

pzalep1:
image

### DIFF
--- a/src/app/cube/usage-stats/heat-map/heat-map.component.scss
+++ b/src/app/cube/usage-stats/heat-map/heat-map.component.scss
@@ -3,7 +3,7 @@
 
   background: white;
   border-radius: 2px;
-  padding: 20px;
+  padding: 40px;
   max-height: 500px;
 
   .title {
--- a/src/app/cube/usage-stats/usage-stats.component.scss
+++ b/src/app/cube/usage-stats/usage-stats.component.scss
@@ -3,7 +3,7 @@
 $row-gap: 25px;
 
 .container {
-  padding: 25px 15px;
+  padding: 30px 50px;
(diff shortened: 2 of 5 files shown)


src/app/cube/usage-stats/heat-map/heat-map.component.scss
```diff
@@ -3,7 +3,7 @@
 
   background: white;
   border-radius: 2px;
-  padding: 20px;
+  padding: 40px;
```
nrwinner: Why would the removal of bootstrap required a doubling of the padding here? Could the root cause of this be the removal of a 'box-sizing' rule?

src/app/cube/usage-stats/usage-stats.component.scss
```diff
@@ -3,7 +3,7 @@
 $row-gap: 25px;
 
 .container {
-  padding: 25px 15px;
+  padding: 30px 50px;
```
nrwinner: Why would the removal of bootstrap required a doubling of the padding here? Could the root cause of this be the removal of a 'box-sizing' rule?

src/app/cube/usage-stats/heat-map/heat-map.component.scss
pzalep1: So it looks like the container has a padding default because it doesn't fill the entire the viewport.

src/app/cube/usage-stats/usage-stats.component.scss
pzalep1: So it looks like the container has a padding default because it doesn't fill the entire the viewport.

PR #878, opened 2019-07-10
Add Page Titles

pzalep1:
The purpose of this PR is to add the ability for our application to dynamically change the page title based on the location of the user within the application. The following is the list of titles for each page: 

Home: **Home | CLARK**
Browse: **Browse Learning Objects | CLARK**
Details: **{{ LO Name }} | CLARK**
Collection Pages: **{{ Collection Name }} | CLARK**
Dashboard: **Your Dashboard | CLARK**
Builder with new LO: **New Learning Object | CLARK**
Builder after name has been added: **{{ LO Name }} | CLARK**
Library: **Your Library | CLARK**
Login: **Login | CLARK**
Register: **Register | CLARK**
Forgot Password: **Change Password | CLARK**
User profiles: **{{ username }} | CLARK** ****This is also for the personal profile
Not Found: **Not Found | CLARK**
Unsupported: **Unsupported | CLARK**
Organization List: **{{ organization name }} | CLARK**

Collection Dashboard
Learning Objects: **Collection Dashboard Learning Objects | CLARK**
Users: **Collection Dashboard Users | CLARK**

### DIFF
--- a/src/app/clark.component.ts
+++ b/src/app/clark.component.ts
@@ -1,9 +1,14 @@
-import { Router } from '@angular/router';
+import { Router , NavigationEnd, ActivatedRoute} from '@angular/router';
 import { Component, OnInit } from '@angular/core';
 import { AuthService } from './core/auth.service';
 import { CartV2Service } from './core/cartv2.service';
 import { ModalService, ModalListElement } from './shared/modals';
 import { trigger, transition, style, animate } from '@angular/animations';
+import { Title } from '@angular/platform-browser';
+import 'rxjs/add/operator/filter';
+import 'rxjs/add/operator/map';
+import 'rxjs/add/operator/mergeMap';
+import { LearningObject } from '@entity';
 
 @Component({
   selector: 'clark-root',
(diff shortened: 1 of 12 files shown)


src/app/clark.component.ts
```diff
@@ -1,9 +1,14 @@
-import { Router } from '@angular/router';
+import { Router , NavigationEnd, ActivatedRoute} from '@angular/router';
 import { Component, OnInit } from '@angular/core';
 import { AuthService } from './core/auth.service';
 import { CartV2Service } from './core/cartv2.service';
 import { ModalService, ModalListElement } from './shared/modals';
 import { trigger, transition, style, animate } from '@angular/animations';
+import { Title } from '@angular/platform-browser';
+import 'rxjs/add/operator/filter';
+import 'rxjs/add/operator/map';
```
nrwinner: Are we using the map and mergeMap imports? If not, we should remove the imports.

pzalep1:
> This looks good to me. I do wonder if we need the `CLARK:` prefix on every page, or just a few. Does the uniqueness of our favicon potentially help people to know that the tab they are seeing is Clark?
> 
> And then the flip side of that is for screen readers, how does tab switching work?

@DonSeannelly I flipped the page name and CLARK so that CLARK is still there for screen readers but is not the first thing that you see visually. The icon is enough for people that are not visually impaired, but the name should stay for the individuals that are.

PR #882, opened 2019-07-12
Deploy Revision Side panel

pzalep1:
***Note: I will remove the dev flag once it's good to go ***
This PR makes some structural changes to the side panel component in the dashboard as well as adds the ability to click the object in the list and open the side panel. 
Screen Shot 2019-07-15 at 3 28 40 PM


src/app/onion/old-dashboard/components/dashboard-item/dashboard-item.component.html
```diff
@@ -1,5 +1,5 @@
 <div *ngIf="!hidden" class="dashboard-item" [ngClass]="{'disabled': disabled, 'shrink': small}">
-  <div class="row-item">
+  <div class="row-item" (click)="actionPermissions('infoPanel')&&viewSidePanel.emit()" [ngClass]="[actionPermissions('infoPanel') ? 'clickable' : 'notClickable']">
```
DonSeannelly: Out of curiosity, since I've been a little out of the loop on 508 things, why is there a click event on a div? Isn't that problematic?

pzalep1: We have a new directive I have to change this when that PR gets merged. The directive will be used in place of the click event so that keyboard and mouseevents will make the function fire.

DonSeannelly: Can we add a FIXME here that notes that?

src/app/onion/old-dashboard/components/side-panel-content/learning-object/learning-object.component.scss
```diff
@@ -1,5 +1,11 @@
 @import '~_vars.scss';
 
+.learning-object {
+  padding: 15px;
+  background: white;
+  border-radius: 4px;
+  box-shadow: 0 0 10px 1px $medium-grey;
```
nrwinner: Can we use an rgba variant of black here? Off the cuff suggestion so play around with it, but using the medium grey shadow on the blue makes the shadow look white at the top.
```suggestion
  box-shadow: 0 0 10px 1px rgba(0, 0, 0, 0.07);
```

PR #886, opened 2019-07-16
Revision Interaction Stubs

pzalep1:
The purpose of this PR is to stub out the interactions that a user will have with the side panel once our revision pipeline is released. This includes the information available in the side panel about the revision, the animations, and the actions that a user can make on a revision.

Screen Shot 2019-07-16 at 3 04 51 PM
Screen Shot 2019-07-16 at 3 05 36 PM 
Screen Shot 2019-07-16 at 3 13 44 PM

**I have a screen capture of the animation. Github won't let me attach it to the PR.**


pzalep1:
> This is feature-toggled off right now, right?

Yes, there is a boolean that is hardcoded false.

pzalep1:
This is feature toggled off with the boolean createRevision in the side-panel-content component

PR #892, opened 2019-07-19
Ignore build-and-test workflow when merging into master

nvisal1:
***Purpose***
This PR is in response to story https://app.clubhouse.io/clarkcan/story/842/only-run-non-deployment-pipeline-on-branches-that-aren-t-master

***Implementation***
Uses the ignore option for the master branch

Circle CI branch filtering docs: https://circleci.com/docs/2.0/configuration-reference/#filters-1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,11 +80,20 @@ workflows:
   version: 2
   build-and-test:
     jobs:
-      - build
-      - test
+      - build:
+          filters:
+            branches:
+              ignore: /master/
+      - test:
+          filters:
+            branches:
+              ignore: /master/
       - publish-image:
           requires:
             - build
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -80,11 +80,20 @@ workflows:
   version: 2
   build-and-test:
     jobs:
-      - build
-      - test
+      - build:
```
gshaw1997: Is it possible to apply a filter to an entire workflow?

gshaw1997: Nvm, doesn't appear to be. :-1: CircleCi

nvisal1: I was looking for that too. Circle CI doesn't seem to support that.

PR #893, opened 2019-07-19
Ignore tag commits in CircleCI

gshaw1997:
**Purpose of this PR**
Addresses https://app.clubhouse.io/clarkcan/story/841/ignore-tags-as-trigger-to-run-workflows-in-circleci.

**Changes in this PR**
- Add filter to ignore tags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,38 +80,57 @@ workflows:
   version: 2
   build-and-test:
     jobs:
-      - build
-      - test
+      - build:
+          filters:
+            tags:
+              ignore: /.*/
+      - test:
+          filters:
+            tags:
+              ignore: /.*/
       - publish-image:
           requires:
             - build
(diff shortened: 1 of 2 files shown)


gshaw1997:
According to CIrcleCi 2.0 docs, tags are ignored unless a filter is specified to trigger a workflow.

"**Executing Workflows for a Git Tag**
CircleCI does not run workflows for tags unless you explicitly specify tag filters. Additionally, if a job requires any other jobs (directly or indirectly), you must use regular expressions to specify tag filters for those jobs. Both lightweight and annotated tags are supported."

https://circleci.com/docs/2.0/workflows/#using-contexts-and-filtering-in-your-workflows

This does not seem to be the case and the config does not support ignoring tags, so seems as though this is a bug with CircleCi and not our config. We should look into opening a support ticket with CircleCi.

PR #902, opened 2019-07-24
Refactor the tooltip directive to use vanilla JS elements instead of angular components

nrwinner:
This PR simplifies the Tooltips module by adding/removing vanilla JS elements to the DOM instead of dynamically rendering tooltip components. It also increases the overall robustness of the tooltips, making it more difficult for them to accidentally appear off-screen. This refactor is part of a larger refactor of the client's shared module.


src/app/admin/admin.module.ts
```diff
@@ -36,14 +36,14 @@ import { LearningObjectListItemComponent } from './components/learning-object-li
     AdminUserCardComponent,
     UserPrivilegesComponent,
     PrivilegesListComponent,
-    LearningObjectListItemComponent
+    LearningObjectListItemComponent,
+    TipDirective
```
DonSeannelly: #MinimizeGitDiff

```suggestion
    TipDirective,
```

PR #904, opened 2019-07-26
Update Revise Button on the Details Page

jbarlow96:
When a learning object is in the review cycle an editor should have the ability to edit the submission. When an editor is viewing a release to learning object and the user does not have a revision created, the editor should have the ability to create a new revision when a learning object is in the review cycle and editor should have the ability to edit the submission. When an editor is viewing a released learning object and the user does not have a revision created, the editor should have the ability to create a new revision.

image

image

**I believe all of the requirements for these two stories are fulfilled, but please double check.**


jbarlow96:
image

image

src/app/onion/learning-object-builder/learning-object-builder.component.ts
```diff
@@ -133,6 +137,18 @@ export class LearningObjectBuilderComponent implements OnInit, OnDestroy {
               // redirect user to dashboard if the object is in the working stage
               if (this.isInReviewStage(learningObject) && !this.authService.hasEditorAccess) {
                 this.router.navigate(['onion/dashboard']);
+              } else if (revision) {
```
pzalep1: I'm not sure if this is the way to do this.

PR #913, opened 2019-07-31
Implement library-service-related error handling

nrwinner:
This PR introduces several small enhancements to the details and library pages.

1) The details page will now disable the 'download' and 'save' buttons if a >= 500 status code is returned from library service. The disabled buttons have a tooltip explaining the situation of a likely-downed library service.

2) The library page now shows an error message and an emoji (for fun) in the event of a likely-downed library service.

3) The download buttons on the Learning Object cards has been removed temporarily. **Note: This functionality should be reimplemented when the Learning Object cards are updated to include graceful error handling**

Screen Shot 2019-07-31 at 2 17 30 PM

Screen Shot 2019-07-31 at 2 17 16 PM


src/app/cube/cart/cart.component.ts
```diff
@@ -21,12 +22,16 @@ export class CartComponent implements OnInit, OnDestroy {
...
+  serviceError: boolean;
+
   statuses = LearningObject.Status;
 
   constructor(
     public cartService: CartV2Service,
     private router: Router,
-    private authService: AuthService
+    private authService: AuthService,
+    private toaster: ToasterService
```
DonSeannelly: #MinimizeGitDiff
```suggestion
    private toaster: ToasterService,
```

src/app/shared/learning-object/learning-object.component.ts
```diff
@@ -32,6 +32,9 @@ export class LearningObjectListingComponent implements OnInit, OnChanges, OnDest
   canDownload = false;
   showDownloadModal = false;
 
+  // FIXME soon pls
+  downloadService = false;
```
DonSeannelly: Can we elaborate on this? What it is/why it is here

PR #916, opened 2019-08-05
New Dashboard 

pzalep1:
This PR addresses some small issues with the dashboard and also removes it from behind the dev flag.


src/app/onion/onion.routing.ts
```diff
@@ -15,19 +13,12 @@ import { OldDashboardComponent } from './old-dashboard/old-dashboard.component';
  *
  * @author Sean Donnelly
  */
-const dashboard = environment.experimental
-  ? {
+const dashboard =  {
```
nrwinner: Now that we're not conditionally routing to different dashboards, can we move this back into the onion_routes object?

nrwinner: I don did it in 6d301f5

PR #923, opened 2019-08-07
Fix bug where back buttons can infinitely point to each other

nrwinner:
Story details: https://app.clubhouse.io/clarkcan/story/952

This PR address ch952 by implementing a snapshot pattern in the HistoryService. When a snapshot is retrieved from the HistoryService, it provides a `rewind` function that navigates to the last location saved in the snapshot and sets the service's history array to the history array in the snapshot.

### DIFF
--- a/src/app/clark.component.ts
+++ b/src/app/clark.component.ts
@@ -67,7 +67,7 @@ export class ClarkComponent implements OnInit {
     private router: Router,
     private activatedRoute: ActivatedRoute,
     private titleService: Title,
-    private _: HistoryService
+    private _: HistoryService,
     ) {
     this.isSupportedBrowser = !(/msie\s|trident\/|edge\//i.test(window.navigator.userAgent));
     !this.isSupportedBrowser ? this.router.navigate(['/unsupported']) :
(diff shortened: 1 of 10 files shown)


src/app/clark.component.ts
```diff
@@ -65,7 +66,8 @@ export class ClarkComponent implements OnInit {
     private cartService: CartV2Service,
     private router: Router,
     private activatedRoute: ActivatedRoute,
-    private titleService: Title
+    private titleService: Title,
+    private _: HistoryService
```
DonSeannelly: image

```suggestion
    private _: HistoryService,
```

src/app/core/history.service.ts
```diff
@@ -1,40 +1,94 @@
...
+
 @Injectable({
   providedIn: 'root'
 })
 export class HistoryService {
 
-  history: NavigationEnd[] = [];
+  private skipNextEvent: boolean;
+
+  history: string[] = [];
```
DonSeannelly: Maybe adding a little JSDoc on this property would help?
```suggestion
  /**
   * A stack containing the URLs of a user's navigation history
   */
  history: string[] = [];
```

src/app/clark.component.ts
nrwinner: image

PR #924, opened 2019-08-07
Space bar in builder outcomes 

pzalep1:
This PR adds logic to the activate directive to ignore an event if it originating from an input.

### DIFF
--- a/src/app/shared/directives/activate.directive.ts
+++ b/src/app/shared/directives/activate.directive.ts
@@ -18,6 +18,10 @@ export class ActivateDirective {
   handleActivate(event: KeyboardEvent | MouseEvent) {
 
     if (event instanceof KeyboardEvent) {
+      // if the active element is an input, disregard the event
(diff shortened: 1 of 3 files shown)


src/app/shared/directives/activate.directive.ts
```diff
@@ -18,6 +18,10 @@ export class ActivateDirective {
   handleActivate(event: KeyboardEvent | MouseEvent) {
 
     if (event instanceof KeyboardEvent) {
+      // if the active element is an input, disregard the event
+      if (document.activeElement.tagName === 'INPUT') {
```
nrwinner: Maybe we should have another clause in this statement to check to make sure the input isn't of type 'checkbox', 'button', 'submit', or 'reset'? Those are all inputs that are technically 'activateable', right?

pzalep1: I'm going to check for the type of text because the directive won't affect any of the other types.

PR #932, opened 2019-08-12
Add alerts back to client 

pzalep1:
The purpose of this PR is to add the ability for the alert banner to be displayed in the client without the need for a deployment.


src/app/components/navbar/components/message/message.component.ts
```diff
@@ -14,30 +15,16 @@ export class MessageComponent implements OnInit {
...
-  toggle() {
-    this.showing = !this.showing;
+    if (environment.production) {
+      setInterval(async () => {
+        this.messages.getStatus().then(message => {
+          this.message = message;
+          this.showing = this.message.isUnderMaintenance;
+        })
+        .catch (e => {
+          // FIXME: Suppress the error
```
nvisal1: What is the reason for this FIXME? Can you add more context to the comment?

src/app/components/navbar/components/message/message.component.ts
```diff
@@ -14,30 +15,16 @@ export class MessageComponent implements OnInit {
...
-    localStorage.setItem('maintenance', this.message.id);
-  }
-
-  toggle() {
-    this.showing = !this.showing;
+    if (environment.production) {
+      setInterval(async () => {
+        this.messages.getStatus().then(message => {
+          this.message = message;
+          this.showing = this.message.isUnderMaintenance;
```
nvisal1: Can you rename `showing` to something like `showBanner` or `displayBanner`?

src/app/components/navbar/components/message/message.component.ts
```diff
@@ -14,30 +15,16 @@ export class MessageComponent implements OnInit {
...
+        })
+        .catch (e => {
```
nrwinner: Since `e` is a throw-away property here, can we rename it to `_`?
```suggestion
        .catch (_ => {
```

src/app/core/messages.service.ts
```diff
@@ -24,11 +24,10 @@ export class MessagesService {
                     catchError(this.handleError)
                 )
                 .toPromise()
-                .then((val: any) => {
-                    if (val && val.length) {
-                        this._message = new Message(val[0]['_id'], 'error', val[0]['maintenanceMessage']);
-                        return this._message;
-                    } else { throw new Error('System status returned a malformed message.'); }
+                .then((val: Message) => {
+                    // tslint:disable-next-line: no-use-before-declare
```
nrwinner: Can we remove this comment and move the below `Message` class to the top of the file?

src/app/components/navbar/components/message/message.component.ts
pzalep1: Resolved with 2c4cee2

src/app/components/navbar/components/message/message.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="wrapper bad" *ngIf="showing">
+<div class="wrapper bad" *ngIf="showBanner === true">
```
nvisal1: Just curious, why doesn't `*ngIf="showBanner"` work?

pzalep1: For some reason it was always evaluating to true even if the variable was set to false.

PR #945, opened 2019-08-22
Design/details page splash

nrwinner:
This PR implements the splash section of the details page. The content is currently hardcoded and will be replaced with Angular goodness when I get the service connected.

Screen Shot 2019-08-22 at 7 59 18 AM


nrwinner:
[ch1027]

PR #960, opened 2019-09-11
Fix bug where unprivileged users are inexplicably granted access to the editor action pad in builder after submission

nrwinner:
This PR fixes a serious bug in the client where users are granted access to the editor action pad immediately following submission, even if the users are not privileged. This granted them the ability to partially bypass our submission process. The user can attempt to release any of their Learning Objects, and while they will show as released in the user's dashboard, the object is never made visible to the public. This presents a worrying loophole wherein the users object is marked as released to the user **and** the collection curators, but invisible any public user, hence putting it in some sort of immutable limbo.

### DIFF
--- a/src/app/onion/shared/submit/submit.component.ts
+++ b/src/app/onion/shared/submit/submit.component.ts
@@ -174,7 +174,6 @@ export class SubmitComponent implements OnInit {
    * @memberof SubmitComponent
    */
   closeModal(submitted?: boolean) {
-    console.log(submitted)
     this.close.emit(submitted || false);
   }
 }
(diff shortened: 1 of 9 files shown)


src/app/onion/shared/submit/submit.component.ts
```diff
@@ -167,7 +173,8 @@ export class SubmitComponent implements OnInit {
    *
    * @memberof SubmitComponent
    */
-  closeModal() {
-    this.close.emit();
+  closeModal(submitted?: boolean) {
+    console.log(submitted)
```
pzalep1: Remove dev log

nvisal1: remove dev log

PR #962, opened 2019-09-17
Change calls to URI Retriever

pzalep1:
The purpose of this PR is to change the calls being made in the Learning Object Builder Store and the Details page to use the new URI Retriever Service.

### DIFF
--- a/src/app/onion/learning-object-builder/builder-store.service.ts
+++ b/src/app/onion/learning-object-builder/builder-store.service.ts
@@ -220,7 +220,7 @@ export class BuilderStore {
 
     return retrieve()
       .then(object => {
-        this.learningObject = object;
+        this.learningObject = object as LearningObject;
         // this learning object is submitted, ensure submission mode is on
         this.validator.submissionMode =
           this.learningObject.status &&
--- a/src/app/onion/learning-object-builder/builder-store.service.ts
+++ b/src/app/onion/learning-object-builder/builder-store.service.ts
@@ -220,7 +220,8 @@ export class BuilderStore {
 
     return retrieve()
       .then(object => {
-        this.learningObject = object as LearningObject;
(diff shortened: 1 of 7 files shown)


src/app/onion/learning-object-builder/builder-store.service.ts
```diff
@@ -212,12 +214,13 @@ export class BuilderStore {
...
     } : async () => {
-      return this.learningObjectService.getLearningObject(id);
+      const value = this.uriRetriever.getLearningObject({id}, ['children', 'parents', 'materials', 'outcomes']);
+      return value.toPromise();
     };
 
     return retrieve()
       .then(object => {
-        this.learningObject = object;
+        this.learningObject = object as LearningObject;
```
nrwinner: Why are we casting `object` to an instance of  `LearningObject` here?

src/app/cube/old-details/details.component.ts
```diff
@@ -163,29 +165,38 @@ export class DetailsComponent implements OnInit, OnDestroy {
...
-            child.status === LearningObject.Status['PROOFING'] ||
-            child.status === LearningObject.Status['WAITING'];
-        }
-      );
+      const resources = ['children', 'parents', 'outcomes', 'materials', 'metrics', 'ratings'];
+      this.uriRetrieverService.getLearningObject({author, name}, resources).pipe(takeUntil(this.isDestroyed$)).subscribe(async (object) => {
+        if (object) {
+          this.releasedLearningObject = object;
+          
+          // FIXME: This filter should be removed when service logic is updated
```
nvisal1: What service logic needs to be updated?

EVogel1999: I don't know, this code was copied from the old service and put into the subscribe of the new service.

PR #971, opened 2019-09-30
Temporarily disable ratings for back-end deployment and toggle message banner

nrwinner:
image


src/app/components/navbar/components/message/message.component.ts
```diff
@@ -15,12 +15,12 @@ export class MessageComponent implements OnInit {
   constructor(private messages: MessagesService) { }
 
   ngOnInit() {
-    if (environment.production) {
+    // if (environment.production) {
```
pzalep1: Why is this commented out?

PR #993, opened 2019-11-07
Remove ratings component on unreleased objects detail pages

pzalep1:
This PR adds a check for the ratings component to see if the object is in any state of being unreleased.


nvisal1:
Looks good. Can you add a check to remove the rating display in the details splash?

PR #1013, opened 2020-01-08
Fix description margin bug

pzalep1:


src/app/cube/details/components/cube-pattern/cube-pattern.component.scss
```diff
@@ -21,7 +21,7 @@
         background: repeating-linear-gradient(white, rgba(255, 255, 255, 0), rgba(255, 255, 255, 0), rgba(255, 255, 255, 0), white);
         position: absolute;
         top: 0;
-        left: 10%;
+        left: 0;
```
pzalep1: This was from before we stretched it out. Should be 0

PR #1015, opened 2020-01-09
Add data for press component

pzalep1:
This PR adds the data needed for the press component. All of the media mentions are currently in a local json file in the assets folder along with the icons of the various news groups that have included us in their articles. The logos might change depending on how they look on the cards.


src/app/core/press-coverage.service.ts
```diff
@@ -0,0 +1,50 @@
...
+      return this.http.get('assets/images/press/mentions.json', { responseType: 'json' })
+        .pipe(
+          retry(3),
+          catchError(this.handleError)
+        )
+        .toPromise()
+        .then((val: Mention[]) => {
+          this._mentions = [];
+          val.forEach(v => {
+            this._mentions.push(new Mention(v.title, v.link, v.icon));
```
EVogel1999: Couldn't we make mentions a interface and say that the http request gets a mentions array?  Also, couldn't we make getMentions() a async function and await the http call?

pzalep1: I made the getMentions function async. Right now this is just retrieving from an internal json store so I think I would like to keep Mention a class with the constructor at the top of this service. If we choose to create a service in the future that has a service we are returning from we can transition to an interface.

PR #1024, opened 2020-01-24
Unclutter footer

EVogel1999:
The purpose of this PR is to help organize and un-clutter the footer so that it is easier to navigate.

Desktop View:
image

Mobile/Tablet View:
image


EVogel1999:
Updated pictures with moved grant info:
image
Mobile:
image

PR #1030, opened 2020-02-03
Add ability to filter by video on browse page

pzalep1:
This PR adds the ability to filter learning objects on the browse page by their inclusion of videos. This PR also adds a title property to the filter so that the verbage of the filters can be better expressed without the need to make large adjustments to how we format the query for Learning Object Service.


pzalep1:
If anyone has any ideas for a better title for this section please drop a comment!

src/app/cube/browse/browse.component.ts
```diff
@@ -59,12 +59,14 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
   filtering = false;
   filters: FilterSection[] = [
     {
+      title: 'collection',
```
nvisal1: Are `name` and `title` synonymous? Why are they set to the same value?

pzalep1: Resolved. I added a comment to the type at the bottom of the file.

PR #1033, opened 2020-02-11
Cube backwards compat

EVogel1999:
Adds route backwards compatibility to the client so old links to content on clark are still valid.  Needs to go in after the gateway PR is merged.

Test URL in development: http://localhost:4200/details/kkuczynski/Principles%20of%20Cyber%20Law%20and%20Policy

### DIFF
--- a/src/app/cube/core/route-backwards-compat.guard.ts
+++ b/src/app/cube/core/route-backwards-compat.guard.ts
@@ -1,7 +1,6 @@
 import { Injectable } from '@angular/core';
 import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree, Router } from '@angular/router';
 import { Observable } from 'rxjs';
-import { router } from 'app/onion/learning-object-builder/components/content-upload/app/content-upload.routes';
 import { HttpClient } from '@angular/common/http';
 import { environment } from '@env/environment';
 
--- a/src/app/cube/core/route-backwards-compat.guard.ts
+++ b/src/app/cube/core/route-backwards-compat.guard.ts
@@ -1,0 +1,15 @@
+import { Injectable } from '@angular/core';
+import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { Observable } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class RouteBackwardsCompatGuard implements CanActivate {
+  canActivate(
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    return true;
+  }
+  
+}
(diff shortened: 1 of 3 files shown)


src/app/cube/core/route-backwards-compat.guard.ts
```diff
@@ -0,0 +1,31 @@
...
+  providedIn: 'root'
+})
+export class RouteBackwardsCompatGuard implements CanActivate {
+
+  constructor(private router: Router, private http: HttpClient) {}
+
+  canActivate(
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+      const cuidRegex = /([a-z0-9]){8}-([a-z0-9]){4}-([a-z0-9]){4}-([a-z0-9]){4}-([a-z0-9]){12}/;
```
nvisal1: We can simply check if the string contains a hyphen instead of performing a complex regex comparison.

src/app/cube/core/route-backwards-compat.guard.ts
```diff
@@ -0,0 +1,31 @@
...
+      if (next.params.learningObjectName.search(cuidRegex) === 0) {
+        return true;
+      } else {
+        return this.http.get(`${environment.apiURL}}/learning-objects/${next.params.username}/${next.params.learningObjectName}`, { withCredentials: false }).toPromise().then(cuid => {
+          return this.router.createUrlTree([`/details/${next.params.username}/${cuid}`]);
+        }).catch(err => {
```
nvisal1: It looks like we are catching and swallowing the error here.
I would like to see an error translation so that we can clearly communicate the error to the user.

src/app/cube/core/route-backwards-compat.guard.ts
EVogel1999: Updated, I made it check to see if the first hyphen in the cuid is where it is supposed to be, i.e. index at 8.

src/app/cube/core/route-backwards-compat.guard.ts
EVogel1999: Added the toaster service to the guard, now if you try to route to something that doesn't make sense like /details/username/demodemo, then you will be rerouted to home with a corresponding error message from the toaster.

src/app/cube/core/route-backwards-compat.guard.ts
```diff
@@ -0,0 +1,32 @@
...
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+      if (next.params.learningObjectName.indexOf('-') === 8) {
+        return true;
+      } else {
+        return this.http.get(`${environment.apiURL}/learning-objects/${encodeURI(next.params.username)}/${encodeURI(next.params.learningObjectName)}`, { withCredentials: true }).toPromise().then(cuid => {
+          return this.router.createUrlTree([`/details/${next.params.username}/${cuid}`]);
+        }).catch(err => {
+          console.error(err);
+          this.toaster.error('Error!', `There was an error getting this learning object's details.`);
```
pzalep1:
```suggestion
          this.toaster.error('Error!', `Unable to get Learning Object details.`);
```

PR #1035, opened 2020-02-18
Fix button not changing when object added to library

pzalep1:
This PR addresses the issue with the library button spinning and being able to be pressed which results in out backend returning an error because the object has already been added to the users library.


src/app/core/library.service.ts
```diff
@@ -28,10 +28,6 @@ export class CartV2Service {
     // this.headers.append('Content-Type', 'application/json');
```
nvisal1: Can you delete this commented code?

PR #1041, opened 2020-02-26
Notifications badge

EVogel1999:
Adds notification badge to the navbar and dropdown.


pzalep1:
> The animation still needs some polishing. The speed is a little slow, some of the cards take too long to fade out, and it snaps into place.

nvisal1:
> The animation still needs some polishing. The speed is a little slow, some of the cards take too long to fade out, and it snaps into place.

I am also still able to trigger the bug where swiping left too fast causes the no notifications display to appear

pzalep1:
> > The animation still needs some polishing. The speed is a little slow, some of the cards take too long to fade out, and it snaps into place.
> 
> I am also still able to trigger the bug where swiping left too fast causes the no notifications display to appear

https://imgflip.com/i/3u2fen

PR #1042, opened 2020-02-28
Move philosophy to bottom

EVogel1999:
Moves the clark philosophy section on the homepage to the bottom of the page so that recently released learning objects appear as the first option on the homepage.

### DIFF
--- a/src/app/cube/home/home.component.scss
+++ b/src/app/cube/home/home.component.scss
@@ -18,7 +18,7 @@
     width: 100%;
     height: 100vw;
     z-index: 1;
-    background: white;
+    background: #f6f9fc;
     transform: skewY(-4deg)
   }
 }
@@ -27,7 +27,6 @@
   position: relative;
   z-index: 1;
   padding: 0 20px;
-  margin-top: -180px;
 
   .about__grid {
     display: grid;
@@ -45,6 +44,7 @@
 }
 
 .learning-objects {
+  margin-top: -220px;
   padding: 80px 20px;
   position: relative;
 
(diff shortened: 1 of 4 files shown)


EVogel1999:
image
image

Olaniyan99:
can we also get a mobile view please.

src/app/cube/home/home.component.scss
```diff
@@ -18,7 +18,7 @@
     width: 100%;
     height: 100vw;
     z-index: 1;
-    background: white;
+    background: #f6f9fc;
```
pzalep1: Do we have a variable for this?

Olaniyan99:
Screen Shot 2020-02-28 at 9 59 56 AM (2)

Also i noticed the "Quality curriculum" module is a little off.

src/app/cube/home/home.component.scss
EVogel1999: No, this is remanent of before we had a global sass variable for almost every color on our site.  I can add it if you want.

EVogel1999:
> Screen Shot 2020-02-28 at 9 59 56 AM (2)
> 
> Also i noticed the "Quality curriculum" module is a little off.

Good catch!  I didn't notice that.  I think that is how it displays if the screen is too zoomed out.  That picture was taken at 25% zoom out I believe.  At 50% on chrome I see this:
image

src/app/cube/home/home.component.scss
pzalep1: Let's use the variables. This makes it easier if we ever need adjust colors for some reason. This way we only have to update the white color we use in one place versus the 100+ stylesheets we have in the client.

PR #1055, opened 2020-03-18
add "objects" to plan c collection page.

Olaniyan99:


src/app/cube/shared/featured/featured.component.html
```diff
@@ -9,6 +9,6 @@
 
 <ng-template #noObjectsTemplate>
   <div class="no-objects">
-    There are no Learning Objects released for this collection. Click <a routerLink="/browse">here</a> to browse all.
+    There are no Learning Objects released for this collection. Click <a routerLink="/browse">here</a> to browse all objects.
```
pzalep1: We also need to remove 'There are no Learning Objects released for this collection.'

Olaniyan99: ohh ok. i'll make those changes asap

PR #1061, opened 2020-03-26
Hotfix/library bug

pzalep1:
This fixes two bugs with the new library. One with mobile where they notifications would cause the side pull out to crash and the spelling of update.

### DIFF
--- a/src/app/cube/library/library.component.html
+++ b/src/app/cube/library/library.component.html
@@ -9,7 +9,7 @@
   
     <div class='library__content-container'>
         <h2 class='library__cards-container__title'>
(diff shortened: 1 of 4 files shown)


src/app/cube/library/library.component.html
```diff
@@ -9,7 +9,7 @@ <h1 class='library__splash-container__title'>
   
     <div class='library__content-container'>
         <h2 class='library__cards-container__title'>
-            {{ notificationCount }} Updates to Items in Your Library
+            {{ notificationCount }} Update<span *ngIf="notificationCount > 1 || notificationCount == 0">s</span> to Items in Your Library
```
nvisal1: This can be simplified to `!== 1`

PR #1063, opened 2020-03-27
Error handling for details page

pzalep1:
This PR adds the 401/403 back to the client and changes the way we are redirecting so we are no longer hitting a route guard for the profile component.


src/app/cube/details/details.component.ts
```diff
@@ -167,32 +167,22 @@ export class DetailsComponent implements OnInit, OnDestroy {
           this.setAcademicLevels();
           this.setLearningObjectAuthors();
           this.loading = false;
-        }
-        this.learningObjectService.fetchLearningObjectWithResources(
```
pzalep1, line 171: This is deleted because Learning Object Service is now updated to send back the revision URI

PR #1066, opened 2020-04-07
Change image back

pzalep1:
This PR changes the client to use the correct docker image. 

This will not pass until the image is updated in docker.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/clark-client
   docker:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/clark-client
   docker:
-    - image: circleci/node:10.13.0
+    - image: cyber4all/circleci-aws:node-alpine
```
pzalep1, line 4:
```suggestion
    - image: cyber4all/circleci-aws:debian
```

PR #1077, opened 2020-04-22
Feature Dashboard PR #2

pzalep1:
This PR adds:
1. The routes for featuring service
2. Featuring Service that:
- Adds ability to retrieve featured
- Adds ability to set featured
- Adds function to retrieve Learning Objects that are not featured

I didn't reuse the Learning Object Service function because I want to keep the component really dumb to any type of filtering logic. I tried something new here with the pattern for the service because of the nature of drag and drop observables are going to be the easiest to deal with here. 

I'm a little stuck on the testing portion, but would like some feedback on the pattern before I spend any more time trying to figure out how to test it effectively.

### DIFF
--- a/src/app/admin/components/draggable-learning-object/draggable-learning-object.component.html
+++ b/src/app/admin/components/draggable-learning-object/draggable-learning-object.component.html
@@ -1,12 +1,12 @@
 <div class="learning-object" [ngClass]="learningObject.length">
     <div class="image">
-        <div class="length">{{ learningObject.length.toUpperCase() }}</div>
+        <div class="length">{{ learningObject?.length.toUpperCase() }}</div>
     </div>
     <div class="learning-object-content">
         <div class="content__topbar">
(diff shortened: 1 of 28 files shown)


src/app/admin/components/draggable-learning-object/draggable-learning-object.component.html
```diff
@@ -1,12 +1,12 @@
...
-        <div class="length">{{ learningObject.length.toUpperCase() }}</div>
+        <div class="length">{{ learningObject?.length.toUpperCase() }}</div>
     </div>
     <div class="learning-object-content">
         <div class="content__topbar">
-            <div *ngIf="collection?.name !== undefined" class="card-collection">{{ collection.name }}</div>
+            <div *ngIf="collection" class="card-collection">{{ collection?.name }}</div>
         </div>
-    </div>
+    </div>
```
EVogel1999, line 9: space?

PR #1082, opened 2020-05-07
Feature Dashboard Release

pzalep1:
This PR changes the featured component on the homepage to retrieve the Featured Learning Objects from the backend.


src/app/cube/shared/featured/featured.component.ts
```diff
@@ -38,7 +39,10 @@ export class FeaturedComponent implements OnInit {
         }
       });
     } else {
-      this.fetchLearningObjects();
+      this.featureService.getFeaturedObjects();
+      this.featureService.featuredObjects.subscribe(objects => {
```
EVogel1999, line 43: This should be unsubscribed when the component is destroyed right?

pzalep1: Yes, good catch.

PR #1089, opened 2020-06-30
Revert "Revert "Initial remapping""

pzalep1:
Reverts Cyber4All/clark-client#1088.

Puts back the outcome deployment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "clark",
   "displayName": "CLARK: Cybersecurity Labs and Resource Knowledge-base",
-  "version": "4.4.1",
+  "version": "4.5.0",
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "clark",
   "displayName": "CLARK: Cybersecurity Labs and Resource Knowledge-base",
-  "version": "4.4.1",
+  "version": "4.5.0",
```
Olaniyan99, line 4: Shouldn’t this be version: 4.5.2?

pzalep1: We're reverting back to the 4.5 and thats 4.4.1 for the last release.

Olaniyan99: ohh right, i read the previous version wrong, thought it said 4.5.1. All good

PR #1092, opened 2020-07-13
Fix Admin dashboard status UI bug

pzalep1:
This PR fixes the bug with the UI status bug. This bug occurred because we were conditionally setting the status button, but not accounting for the context menu being created at the bottom of the page and it was unable to find it's anchor element resulting in an error.


Olaniyan99:
looks good

PR #1101, opened 2020-09-14
Fix password check error

pzalep1:
There is logic in the client to check for strong passwords, but due to an update from Angular 7 from early 2019 the ViewChild was returning undefined and the form was being allowed to pass without any validation being done on client. In order to get this to work I added { static: true } which resolves query results before change detection runs which now properly executes the validation.


EVogel1999 (approved):
Looks good!

Olaniyan99 (approved):

davidnso (approved):

PR #1105, opened 2020-11-03
Update change authorship

pzalep1:
This PR updates the route and service call to reflect the adjustment made to the backend. It also updates the message to the user to include a notice of revisions of the object and revisions of its children having their author changed. 
Screen Shot 2020-11-03 at 1 25 43 PM

### DIFF
--- a/src/app/admin/core/authorship.service.ts
+++ b/src/app/admin/core/authorship.service.ts
@@ -16,11 +16,9 @@ export class AuthorshipService {
 
   async changeAuthorship(oldAuthor: User, id: string, newAuthor: string) {
     return this.http
-    .post(ADMIN_ROUTES.CHANGE_AUTHOR(oldAuthor.username, id),
+    .post(ADMIN_ROUTES.CHANGE_AUTHOR(oldAuthor.id, id),
     {
-      'fromUserID': oldAuthor.id,
-      'toUserID': newAuthor,
-      'objectID': id,
+      'author': newAuthor,
     },
     { withCredentials: true, responseType: 'text'}
     )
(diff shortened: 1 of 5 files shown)


src/app/admin/core/authorship.service.ts
```diff
@@ -16,11 +16,9 @@ export class AuthorshipService {
 
   async changeAuthorship(oldAuthor: User, id: string, newAuthor: string) {
     return this.http
-    .post(ADMIN_ROUTES.CHANGE_AUTHOR(oldAuthor.username, id),
+    .post(ADMIN_ROUTES.CHANGE_AUTHOR(oldAuthor.id, id),
     {
-      'fromUserID': oldAuthor.id,
-      'toUserID': newAuthor,
-      'objectID': id,
+      'author': newAuthor,
```
mfranz1, line 21: So this post updates the author  to newAuthor, for the removed author, shouldn't there be some update that removes this object from that user being the author?

src/environments/route.ts
```diff
@@ -21,8 +21,8 @@ export const ADMIN_ROUTES = {
   REMOVE_MAPPER(userId: string): string {
     return `${environment.apiURL}/guidelines/members/${encodeURIComponent(userId)}`;
   },
-  CHANGE_AUTHOR(userId: string, cuid: string): string {
-    return `${environment.apiURL}/admin/users/${encodeURIComponent(userId)}/learning-objects/${cuid}/change-author`;
+  CHANGE_AUTHOR(userId: string, id: string): string {
```
mfranz1, line 24: I think this answers my previous question

Cwagne17 (approved):
Makes sense, changing description, and function.  Looks good.

EVogel1999 (approved):

src/app/admin/core/authorship.service.ts
pzalep1: @mfranz1 So that is done completely on the backend. When we switch an author we are switching who has the ability to edit it in an unreleased state.

PR #1107, opened 2020-11-09
:construction_worker: Add Husky Pre-Commit Hooks for Lint and Test

DonSeannelly:
This PR adds a pre-commit hook that will run linting and then unit tests before finalizing the commit. In order to get this work, there were several linting errors fixed across files.


src/app/onion/learning-object-builder/builder-store.service.ts
```diff
@@ -969,7 +969,7 @@ export class BuilderStore {
     if (updateValue.serviceId) {
       updateValue.id = updateValue.serviceId;
     }
-    console.log('outcome',outcome);
+    console.log('outcome', outcome);
```
pzalep1, line 972: This could just get :fire:.

src/app/shared/modules/filesystem/DirectoryNode.ts
```diff
@@ -0,0 +1,185 @@
+import { LearningObject } from '@entity';
+
+/**
+ * Class representing simple Node in DirectoryTree
```
pzalep1, line 4: 👍

src/app/cube/user-profile/user-preferences/user-preferences.component.html
```diff
@@ -1,5 +1,5 @@
-<!-- <app-contextmenu [content]="modalCtrl.contextMenuContent"></app-contextmenu>
-<app-dialogmenu [content]="modalCtrl.dialogMenuContent"></app-dialogmenu> -->
+<!-- <clark-contextmenu [content]="modalCtrl.contextMenuContent"></clark-contextmenu>
+<clark-dialogmenu [content]="modalCtrl.dialogMenuContent"></clark-dialogmenu> -->
```
EVogel1999, line 2: If this is commented out then shouldn't it be deleted?

DonSeannelly: Good catch! I didn't notice this one because I did a bulk find and replace

DonSeannelly: This component looked a little odd to me, so I looked into it a little more and it seems that it was never fully built out:
image

@pzalep1 it might be a good idea for this to be removed in a future ticket

pzalep1 (approved):

EVogel1999 (approved):

PR #1124, opened 2021-01-22
Fix guidelines in builder

pzalep1:
The builder was not parsing a StandardOutcome correctly after a learning object had been submitted to a collection and then pulled back from the waiting state.

### DIFF
--- a/src/app/onion/learning-object-builder/components/standard-outcomes/standard-outcomes.component.ts
+++ b/src/app/onion/learning-object-builder/components/standard-outcomes/standard-outcomes.component.ts
@@ -110,6 +110,7 @@ export class StandardOutcomesComponent implements OnChanges, OnDestroy {
   }
 
   toggleStandardOutcome(standardOutcome: StandardOutcome, value: boolean) {
+    standardOutcome = new StandardOutcome(standardOutcome);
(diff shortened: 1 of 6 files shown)


src/app/onion/learning-object-builder/components/standard-outcomes/standard-outcomes.component.ts
```diff
@@ -110,6 +110,7 @@ export class StandardOutcomesComponent implements OnChanges, OnDestroy {
   }
 
   toggleStandardOutcome(standardOutcome: StandardOutcome, value: boolean) {
+    standardOutcome = new StandardOutcome(standardOutcome);
```
mfranz1, line 113: What is happening here? Why are you creating a standardOutcome from the standardOutcome that was passed to the toggle function?

mfranz1 (approved):

src/app/onion/learning-object-builder/components/standard-outcomes/standard-outcomes.component.ts
pzalep1: The StandardOutcome is not typed correctly in parameter. It is just an object. There appears to be a larger issue with the StandardOutcome type in the client, but it will take some time to figure out.

PR #1125, opened 2021-01-22
Paigezaleppa/about us

pzalep1:
This PR adds a small About Us section to the homepage with a button that will open a modal that will play our new video. 

Screen Shot 2021-01-28 at 9 58 14 AM

Screen Shot 2021-01-28 at 10 24 05 AM


mfranz1 (approved):
Looks good; only comment I have is we host all this learning curriculum but we can't host our own video?
(just joking)

src/app/cube/home/components/what-clark/what-clark.component.html
```diff
@@ -0,0 +1,34 @@
...
+      </div>
+      Securing the nation through education
+    </div>
+    <div class="what__description">
+      {{ copy.WHAT_BLURB }}
+      <a routerLink="/about">Learn more</a>
+    </div>
+    <div class="what__actions">
+        <div class="about__link">
+            <div role=button class="play-button" (activate)="openModal()" aria-label="Play CLARK Video"><i class="fal fa-play-circle"></i><span class="button-text">Watch our video</span></div>
```
EVogel1999, line 14: Why is this a div acting like a button?

EVogel1999 (changes requested):
One small thing.

src/app/cube/home/components/what-clark/what-clark.component.html
pzalep1: I've learned that it's okay to have a div act as a button. The role property will take care of the screen reader for us which is actually really cool.

EVogel1999 (approved):

devky690 (approved):

PR #1134, opened 2021-03-02
Cookie Popup and Terms of Service

Cwagne17:
PR completes [Story 3640](https://app.clubhouse.io/clarkcan/story/3640/add-better-description-and-link-to-terms-of-service-to-cookie-agreement)

Banner was updated to include terms of service.  New cookie popup appears in auth module when banner is declined (when popup is visible the buttons will not work).  The terms of service is updated as well with SecurEd inc. and Cookie Terms of Service section.

image
image
image


src/app/core/cookie-agreement.service.ts
```diff
@@ -0,0 +1,52 @@
...
+   * @Check_Functions
+   * checks local storage for value of
+   */
+  private checkCookieAgreement() {
+    const agreement = localStorage.getItem('acceptCookieAgreement');
+    this.cookiesAgreement = agreement !== null && agreement === 'true';
+  }
+  private checkShowCookieBanner() {
+    const agreement = localStorage.getItem('showCookieBanner');
+    this.showCookieBanner = !(agreement !== null && agreement === 'false');
```
pzalep1, line 32: Since this in production, what if I have accepted the cookie agreement in the past? Will there be any issue with me logging in? If there is we should make sure to set the show cookieBanner appropriately. We haven't updated our terms of service so if they have already accepted it we can allow them to login.

mfranz1 (approved):

src/app/shared/components/cookie-popup/cookie-popup.component.html
```diff
@@ -0,0 +1,23 @@
+<div class="wrapper">
+  <div class="container">
+    <p>
+      Must accept cookies outlined in the
```
pzalep1, line 4:
```suggestion
      You must accept cookies outlined in the
```

src/app/shared/components/cookie-popup/cookie-popup.component.scss
```diff
@@ -0,0 +1,29 @@
...
+      appearance: none;
+      border-radius: 4px;
+      border: none;
+      padding: 12px 25px;
+      text-transform: uppercase;
+      cursor: pointer;
+    }
+  }
+}
+@media screen and (max-width: 500px) {
```
pzalep1, line 26: If this is empty it should be removed.

pzalep1 (changes requested):

src/app/components/cookies/cookies.component.scss
```diff
@@ -47,15 +38,31 @@
 
         .text {
             font-size: $normal;
+            a {
+                color: white;
```
EVogel1999, line 42: Can we add an underline to the link so we know it's there?

src/app/core/cookie-agreement.service.ts
```diff
@@ -0,0 +1,52 @@
...
+    return this.cookiesAgreement;
+  }
+
+  getShowCookieBannerVal(): boolean {
+    this.checkShowCookieBanner();
+    return this.showCookieBanner;
+  }
+
+  /**
+   * @Check_Functions
```
EVogel1999, line 23: What does this do?

src/app/cube/terms-of-service/terms-of-service.component.html
```diff
@@ -91,6 +91,46 @@ <h3 tabindex="0">Changes in Service</h3>
       without providing prior notice. We also retain the right to create limits on your use of the Service
       including storage, at our sole discretion, at any time without prior notice to you.</p>
 
+    <h3>Cookies</h3>
+    <p>
+      We use cookies and other technologies on our application for functionality
+      and performance. The first use of cookies on CLARK is for authorization
+      purposes, this allows us to refer to your account information quick and
+      easily. This is also what makes it possible for you to browse our
```
EVogel1999, line 99: 'this allows us to access your CLARK account information quickly and easily'.

src/app/cube/terms-of-service/terms-of-service.component.html
```diff
@@ -91,6 +91,46 @@ <h3 tabindex="0">Changes in Service</h3>
...
+      application without having to login between each refresh or new page. The
```
EVogel1999, line 100: 'between each page refresh or navigating our site'

src/app/cube/terms-of-service/terms-of-service.component.html
```diff
@@ -91,6 +91,46 @@ <h3 tabindex="0">Changes in Service</h3>
...
+      cookie stored is created by us and no third party is used, the information
```
EVogel1999, line 101: I would remove the "and no third part is used"

EVogel1999 (changes requested):
Looking good!  Just have a few other suggestions in addition to Paige's comments.

src/app/core/cookie-agreement.service.ts
Cwagne17: @pzalep1 I did update the terms of service to add the cookies paragraph similar to CARD. With current implementation there will be a problem requiring them to accept again.  However, this can be avoided by changing  acceptCookieAgreement (new) to cookieAgreement (currently) as the localStorage value to look for.

pzalep1: That sounds good. By updated I we didn't "add" anything new we just added further explanation on our cookies.

src/app/cube/terms-of-service/terms-of-service.component.html
pzalep1: We need to make this explicit so it should be left.

src/app/core/cookie-agreement.service.ts
Cwagne17: I just changed it to display a comment that explains the checking functions purpose
"The checking functions are helper functions to check the values stored in users localStorage."

src/app/core/cookie-agreement.service.ts
Cwagne17: Ok sounds good. I just added logic and tested it. Anyone who has previously accepted the cookies will not be requested to accept again. This works for both the banner and the popup. @pzalep1

pzalep1 (approved):
Looks good

devky690 (approved):

jmulholland4014 (approved):

PR #1144, opened 2021-04-16
3898/icon-in-footer-overlaps-with-text

EVogel1999:
This fixes the footer issue between switching from desktop to mobile where the logos would appear off-screen or overlap text.  Also includes a small quality of life dockerfile upgrade so if you are running the client docker container you can go directly to links instead of having to go to the homepage first and work your way to the link.  Completes story [3898/icon-in-footer-overlaps-with-text](https://app.clubhouse.io/clarkcan/story/3898/icon-in-footer-overlaps-with-text).

### DIFF
--- a/nginx-custom.conf
+++ b/nginx-custom.conf
@@ -1,0 +1,9 @@
+server {
+    listen 80;
+    server_name  localhost;
+    location / {
+        root   /usr/share/nginx/html/;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;
+    }
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,4 +23,5 @@ FROM nginx:alpine as serve
 # Copy the build folder from the build stage into the nginx folder
 # and expose the port
 COPY --from=build /build/dist /usr/share/nginx/html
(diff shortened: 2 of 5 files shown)


EVogel1999:
Screen Shot 2021-04-16 at 9 50 44 AM

Cwagne17 (approved):

nginx-custom.conf
```diff
@@ -0,0 +1,9 @@
+server {
+    listen 80;
+    server_name  localhost;
+    location / {
```
pzalep1, line 4: What are these changes for?

devky690 (approved):

nginx-custom.conf
EVogel1999: This has to do with the quality of life dockerfile upgrade.  Basically if you have a localhost link and you copy it and try to go to it in another tab, it will say it cannot resolve the host so you are forced to manually go through the site again to where the link takes you.  But this allows you to just go directly to the location via the url

mfranz1 (approved):

pzalep1 (approved):

PR #1187, opened 2021-09-09
Stringify the query

pzalep1:
This PR allows for the limit and page to be set correctly.

### DIFF
--- a/src/environments/route.ts
+++ b/src/environments/route.ts
@@ -587,13 +587,9 @@ export const STANDARD_GUIDELINE_ROUTES = {
     limit?: string,
     type?: string
   }) {
-    return `${environment.apiURL}/frameworks?
-      text=${encodeURIComponent(params.text)}&
-      year=${encodeURIComponent(params.year)}&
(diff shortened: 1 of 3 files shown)


src/environments/route.ts
```diff
@@ -587,13 +587,9 @@ export const STANDARD_GUIDELINE_ROUTES = {
...
-      text=${encodeURIComponent(params.text)}&
-      year=${encodeURIComponent(params.year)}&
-      levels=${encodeURIComponent(params.levels)}&
-      page=${encodeURIComponent(params.page)}&
-      limit=${encodeURIComponent(params.limit)}&
-      type=${encodeURIComponent(params.type)}`;
+    const wow = `${environment.apiURL}/frameworks?${querystring.stringify(params)}`;
+    console.log(wow);
+    return wow;
   },
```
EVogel1999, line 593: Should this have a console log and the variable to be called 'wow'?

EVogel1999 (approved):

PR #1191, opened 2021-09-12
Add ability to delete a revision from admin dashboard

pzalep1:
This PR closes the frontend tasks in: https://app.shortcut.com/clarkcan/story/5043/update-frontend-and-backend-to-handle-multiple-revision-error

New admin options for learning objects: 
Released with a revision in waiting, review, proofing:
Screen Shot 2021-09-12 at 1 00 25 PM

Released with no revision:
Screen Shot 2021-09-12 at 1 06 00 PM

Waiting, review, or proofing:
Screen Shot 2021-09-12 at 1 06 09 PM


mfranz1 (approved):

EVogel1999 (approved):

jouypo23 (approved):

src/app/admin/core/unrelease.service.ts
```diff
@@ -33,6 +33,25 @@ export class UnreleaseService {
       .toPromise();
   }
 
+  /**
+   * Deletes a revision of a learning object. This is designed to allow an editor to create a new
```
StarChaser44, line 37: I don't know if this is just the way I am seeing it, but you could make combine the first two comments together because they kind of look broken apart. For example, you could try this:

* Deletes a revision of a learning object. This is designed to allow an editor to create a new revision when it is necessary for the editorial process to continue.

OR
* Deletes a revision of a learning object. This is designed to allow an editor
* to create a new revision when it is necessary for the editorial process to continue.

src/entity/learning-object/learning-object.ts
```diff
@@ -233,6 +233,14 @@ export class LearningObject {
     this._revisionUri = val;
   }
 
+  get revision(): any {
```
StarChaser44, line 236: Can we possibly make this an interface?

StarChaser44 (approved):
This looks very nice. One thing I am wondering is if there should be coding standards for consistency among all devs.

Cwagne17 (approved):

PR #1227, opened 2021-12-14
:construction: Added newsletter popup component :art:

mfranz1:
This PR completes shortcut story: https://app.shortcut.com/clarkcan/story/6348/add-pop-up-to-sign-up-for-clark-newsletter and also removes the unused ChangeAuthorshipComponent from the shared components folder. 

A popup is now generated when a user hits clark on their first visit (or after cookies are cleared). The popup behaves the same as how cookie notifications work. The subscription handling is done via 3rd party through sendgrid. The component on clark uses an HTML iframe element to interact with the sendgrid form. We toggle the interaction with the popup by using a localStorage boolean 'showSubscriptionBanner'. The flow is for a user to subscribe and then either click off the popup or close the popup. The popup will remain hidden until cookies are cleared or it is triggered in the footer.

image

image

image


src/app/cube/shared/footer/footer.copy.ts
```diff
@@ -10,5 +10,6 @@ export const COPY = {
     ACCESSIBILITY: 'Accessibility',
     PRESS: 'Press and Media',
     OUTAGES: 'System Outages',
-    ABOUT: 'About Us'
+    ABOUT: 'About Us',
+    SUBSCRIBE: 'Newsletter Sign Up'
```
pzalep1, line 14:
```suggestion
    SUBSCRIBE: 'Sign up for our Newsletter'
```

src/app/core/subscription-agreement.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class SubscriptionAgreementService {
+  showSubscriptionBanner = true;
+
+  constructor() {}
+
+  // Service call to return current value of subscription banner toggle
```
pzalep1, line 11:
```suggestion
  // Local storage call to return current value of subscription banner toggle
```

src/app/core/subscription-agreement.service.ts
```diff
@@ -0,0 +1,31 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class SubscriptionAgreementService {
```
pzalep1, line 6: Can we add a JS doc to this class to explain that this subscription is for the newsletter and not our subscription model?

src/app/components/subscription/subscription.component.html
```diff
@@ -0,0 +1,2 @@
+<!--Sendgrid element-->
+<iframe class="clark-subscription" src="https://cdn.forms-content.sg-form.com/9a369384-51e4-11ec-a0b6-92cdb9380ecd"></iframe>
```
pzalep1, line 2: Can you put a comment to the docs from SendGrid?

src/app/clark.component.ts
```diff
@@ -92,10 +93,6 @@ export class ClarkComponent implements OnInit {
         }
       });
 
-    if (localStorage.getItem('cookieAgreement')) {
```
pzalep1, line 95: Why was this removed?

pzalep1 (changes requested):
Double check that component you deleted and I made a wording suggestion. I like using the service for this. Keeps it clean in the clark component and other components you're using it in

src/app/clark.component.ts
mfranz1: Everything relating to cookies is now handled in CookieAgreementService so this didn't serve a purpose anymore.

src/app/core/subscription-agreement.service.ts
```diff
@@ -0,0 +1,51 @@
...
+  showSubscriptionBanner = true;
+
+  constructor() {}
+
+  /**
+   * Function that toggles the value for the subscription banner
```
mfranz1, line 19:
```suggestion
   * Function that retrieves the toggle value for the subscription banner
```

src/app/clark.component.scss
```diff
@@ -4,4 +4,9 @@ clark-cookies {
     bottom: 0;
     left: 0;
     right: 0;
+}
+
+.modal-container {
+    height: 475px;
+    width: 560px;
 }
```
EVogel1999, line 12: Will this work on mobile?

EVogel1999 (commented):
Just one comment so far.

pzalep1 (approved):

src/app/clark.component.scss
mfranz1: @EVogel1999 this is not mobile friendly, will need to look into sendgrid docs to adjust their element sizing on our page.

jouypo23 (approved):
Want to double-check the newsletter signup link is at the bottom of the general information column footer, but other than that it looks good.

src/app/core/subscription-agreement.service.ts
```diff
@@ -0,0 +1,51 @@
...
+   * @const banner checks local storage for 'showSubscriptionBanner'
+   * If null: intialize it^ @function setShowSubscriptionBanner(true)
+   */
+  private checkShowSubscriptionBanner() {
+    const banner = localStorage.getItem('showSubscriptionBanner');
+    if (banner == null) {
+      this.setShowSubscriptionBanner(true);
+    }
+    this.showSubscriptionBanner = !(banner !== null && banner === 'false');
+  }
```
JosueMolinaMorales, line 40: Is there a reason why you call this.setShowSubscriptionBanner() in the in but not outside? just curious.

src/app/clark.component.scss
StarChaser44: Nice context

StarChaser44 (approved):

src/app/core/subscription-agreement.service.ts
mfranz1: @JosueMolinaMorales this if statement helps to create the local storage variable if it did not already exist. The private function 'checkShowSubscriptionBanner' is hit every time we call this class so we don't want to continuously initialize the localVariable to true

Cwagne17 (approved):

EVogel1999 (approved):

pzalep1 (approved):

PR #1276, opened 2022-03-30
7912/child object does not delete from a hirearchy

EVogel1999:
This PR adds a condition to remove a learning object child from a parent learning object.  It uses the new PATCH route to accomplish this.  Completes story [7912/child-object-does-not-delete-from-a-hirearchy](https://app.shortcut.com/clarkcan/story/7912/child-object-does-not-delete-from-a-hirearchy).


mfranz1 (approved):

src/app/onion/core/learning-object.service.ts
```diff
@@ -434,21 +434,36 @@ export class LearningObjectService {
   setChildren(
     learningObjectName: string,
     authorUsername: string,
-    children: string[]
+    children: string[],
```
pzalep1, line 437: If remove is true these are the children being removed right?

EVogel1999: Yes.  But there is really only one child removed at a time because there is a confirmation popup each time you remove a child

pzalep1 (approved):

PR #1322, opened 2022-05-31
fixed hidden LO submition button bug

AStoerkel1:
this PR is related to [8718](https://app.shortcut.com/clarkcan/story/8718/clark-submission-button-size-issue) in shortcut


src/app/onion/shared/submit/submit.component.scss
```diff
@@ -2,9 +2,10 @@
 
 .submit {
   max-width: 800px;
-  min-height: 550px;
+  height: 85vh;
```
EVogel1999, line 5: I think we should use a mix of ```max-height``` and ```height``` here instead.  That way the popup isn't too big on larger screens.

EVogel1999 (commented):
One suggestion.

src/app/onion/shared/submit/submit.component.scss
AStoerkel1: I can give it a shot but the popup doesn't really do what you would expect it to do, even set to 85vh it is still kinda cut off on the top and bottom when zoomed in far enough to experience the bug.

AStoerkel1: ok I added a max-height of 550px

EVogel1999 (approved):
Looks good!  Thank you!

pzalep1 (approved):

PR #1332, opened 2022-06-06
Added released status to browse page query

StarChaser44:
Completes story [11417](https://app.shortcut.com/clarkcan/story/11417/display-released-learning-objects-only-on-the-browse-page)

### DIFF
--- a/src/app/cube/browse/browse.component.ts
+++ b/src/app/cube/browse/browse.component.ts
@@ -37,6 +37,7 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
     sortType: undefined,
     collection: '',
     fileTypes: [],
+    status: [LearningObject.Status.RELEASED]
   };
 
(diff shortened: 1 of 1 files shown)


AStoerkel1 (approved):

src/app/cube/browse/browse.component.ts
```diff
@@ -37,6 +37,7 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
     sortType: undefined,
     collection: '',
     fileTypes: [],
+    status: [LearningObject.Status.RELEASED]
```
mfranz1, line 40: This still works even without a text query? (I.e. just going to the browse page from the homepage with no query)

StarChaser44: I was able to confirm that it does correctly load only released learning objects when there is no text query

PR #1353, opened 2022-06-14
7126/new login

AStoerkel1:
this PR completes [7126](https://app.shortcut.com/clarkcan/story/7126/implement-new-login-experience)
- as of jun 14 3:30pm still not able to get the SSO route to redirect through the client, It makes the request to clark-gateway but the request never gets to user-service. it just throws a 401: Unauthorized error. 
- localhost:4000/google (user-service route) takes me to the SSO page and I can log in there, then when I go back to click the SSO button again, it gives the error pictured below

Screen Shot 2022-06-14 at 3 25 03 PM
 
UPDATE:
- fixed this bug by making it a link, and not calling a method from auth service


src/app/new-auth/login/login.component.ts
```diff
@@ -1,23 +1,121 @@
...
+    private authValidation: AuthValidationService,
+    private route: ActivatedRoute,
+    private auth: AuthService,
+    private router: Router
+    ) {
+      this.route.parent.data.subscribe(() => {
+        if (route.snapshot.queryParams.redirectUrl) {
+          this.redirectUrl = decodeURIComponent(route.snapshot.queryParams.redirectUrl);
+        }
+      });
```
EVogel1999, line 59: You'll also want to check if there is an err message for single sign on to display to the user.

EVogel1999 (commented):
One small update having to do with the user-service PR you put in.

src/app/new-auth/login/login.component.ts
pzalep1: ^

src/app/new-auth/login/login.component.ts
```diff
@@ -1,23 +1,121 @@
...
+    this.authInfo = form.value;
+    //get form control for all fields
+    const pwordFormCtl = this.password.valueAccessor.control;
+    const usernameFormCtl = this.username.valueAccessor.control;
+
+    if(usernameFormCtl.hasError('required') || pwordFormCtl.hasError('required')) {
+      this.bannerMsg = 'Please fill out all required fields';
+      this.authValidation.showError();
+    } else if (this.attempts < 5){
+      this.attempts++;
```
pzalep1, line 82: If someone leaves their page open for an hour to wait does this reset? I know it sounds crazy, but just wondering.

pzalep1 (commented):
Just the one question from me.

src/app/new-auth/login/login.component.ts
AStoerkel1: thats a good point, it also isn't persistent if the browser is closed, like if someone maxes out their attempts, quits and tries again in 45min it'll be back at 0. I can figure out the first problem but the second one seems tricky if it can even be done on the client.

pzalep1: That's a session implementation that we haven't gotten to on the backend.

src/app/new-auth/login/login.component.ts
AStoerkel1: Ok so I added displaying the error message on the front end, but, when i try to test it with an account that isn't registered in the database, it doesn't redirect correctly. it's almost like when an error happens it doesn't hit the catch in handleGoogleRedirect(). it's handled somewhere else.

pzalep1: Check to see if there is an errorParser or errorHandler in a catch somewhere. Some of them are setup to throw out a bunch of errors

AStoerkel1:
Screen Shot 2022-06-21 at 1 01 27 PM

src/app/new-auth/login/login.component.ts
```diff
@@ -1,23 +1,149 @@
...
+  authInfo: {username: string, password: string};
+  errorMsg = '';
+  bannerMsg = '';
+  attempts = 0;
+  @ViewChild('username')
+  username;
+  @ViewChild('password')
+  password;
+  redirectUrl;
+  gatewayUrl = environment.apiURL;
```
JosueMolinaMorales, line 51: For better readability:
```suggestion
  finalAttempt: number;
  errorMsg = '';
  bannerMsg = '';
  attempts = 0;
  loginFailure: Boolean = false;
  isNameLogin = false;
  
  authInfo: {
      username: string, 
      password: string
  };
  
  @ViewChild('username')
  username;
  @ViewChild('password')
  password;
  
  redirectUrl;
  gatewayUrl = environment.apiURL;
```

JosueMolinaMorales (commented):
Just one small thing

PR #1381, opened 2022-07-15
Andy/material notes

AStoerkel1:
This PR Completes [11612](https://app.shortcut.com/clarkcan/story/11612/add-material-notes-to-builder) on shortcut.
Screen Shot 2022-07-21 at 11 17 12 AM

Screen Shot 2022-07-21 at 11 11 35 AM

### DIFF
--- a/src/app/admin/pages/learning-objects/learning-objects.component.ts
+++ b/src/app/admin/pages/learning-objects/learning-objects.component.ts
@@ -156,7 +156,7 @@ export class LearningObjectsComponent
         relativeTo: this.route,
         replaceUrl: true
       }
-    )
+    );
   }
 
   /**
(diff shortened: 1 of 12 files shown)


JosueMolinaMorales (approved):

mfranz1 (approved):

pzalep1:
Could you add ^ and the inverse to the right hand side of the options so it looks more like an accordian? I would not think that they open up like that at first glance.

pzalep1 (commented):
Code looks good. Just the comment I already left and:
1. Could we get a screenshot of the whole page now that the middle column doesn't stretch all the way?
2. Could you add just a tiny bit more padding around each option in the list?

pmanal1 (approved):

PR #1417, opened 2022-08-04
Andy/help collections link

AStoerkel1:
this PR completes [12407](https://app.shortcut.com/clarkcan/story/12407/implement-help-explore-collections-section)

It just links to the collections page

### DIFF
--- a/src/app/cube/new-home/help/components/help-card/help-card.component.html
+++ b/src/app/cube/new-home/help/components/help-card/help-card.component.html
@@ -1,8 +1,8 @@
-<div class="helpCard">
+<div class="helpCard" [routerLink]="option.link">
     <span [ngClass]="option.iconColor">
         <i [ngClass]="option.icon"></i>
     </span>
     <h2>{{ option.title }}</h2>
     <p>{{ option.description }}</p>
-    <a>Get Started <i class="fa fa-angle-right" aria-hidden="true"></i></a>
+    <a [routerLink]="option.link">Get Started <i class="fa fa-angle-right" aria-hidden="true"></i></a>
 </div>
--- a/src/app/cube/new-home/help/components/help-card/help-card.component.ts
+++ b/src/app/cube/new-home/help/components/help-card/help-card.component.ts
@@ -7,7 +7,7 @@ import { Component, Input, OnInit } from '@angular/core';
 })
 export class HelpCardComponent implements OnInit {
 
-  @Input() option: {title: string, description: string, icon: string, iconColor: string};
+  @Input() option: {title: string, description: string, icon: string, iconColor: string, link?: string[]};
 
   constructor() { }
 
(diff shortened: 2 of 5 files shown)


src/app/cube/new-home/help/components/help-card/help-card.component.html
```diff
@@ -1,8 +1,8 @@
-<div class="helpCard">
+<div class="helpCard" [routerLink]="option.link">
     <span [ngClass]="option.iconColor">
         <i [ngClass]="option.icon"></i>
     </span>
     <h2>{{ option.title }}</h2>
     <p>{{ option.description }}</p>
-    <a>Get Started <i class="fa fa-angle-right" aria-hidden="true"></i></a>
+    <a [routerLink]="option.link">Get Started <i class="fa fa-angle-right" aria-hidden="true"></i></a>
```
JosueMolinaMorales, line 7: what happens when option.link is undefined since its an optional field?

JosueMolinaMorales (commented):
Just a question

src/app/cube/new-home/help/components/help-card/help-card.component.html
AStoerkel1: The component doesn't link anywhere and Emily's template implementation still works.

src/app/cube/new-home/help/components/help-card/help-card.component.ts
```diff
@@ -7,7 +7,7 @@ import { Component, Input, OnInit } from '@angular/core';
 })
 export class HelpCardComponent implements OnInit {
 
-  @Input() option: {title: string, description: string, icon: string, iconColor: string};
+  @Input() option: {title: string, description: string, icon: string, iconColor: string, link?: string[]};
```
EVogel1999, line 10: I don't think link should be a string array.

EVogel1999 (commented):
One small suggestion about the link being an array

src/app/cube/new-home/help/components/help-card/help-card.component.ts
AStoerkel1: I made it that way so that i could use it as is in the routerlink. what would you recommend?

src/app/cube/new-home/help/components/help-card/help-card.component.html
pzalep1: Is there ever a case where option could be undefined?

pzalep1 (commented):
Just the one question.

src/app/cube/new-home/help/components/help-card/help-card.component.html
AStoerkel1: all of the options are hard-coded in the help component and emily used an ngFor to iterate over them, unless we add an empty object to the array then I don't think so.

pzalep1 (approved):

JosueMolinaMorales (approved):
Looks good!

mfranz1 (approved):

EVogel1999 (approved):

PR #1419, opened 2022-08-08
Andy/collections section

AStoerkel1:
### The Text reads "Learning Objects are grouped into Collections within CLARK. Each collection is unique, has its own peer-review process, and is supported by its own community."
Screen Shot 2022-08-09 at 10 08 30 AM
Screen Shot 2022-08-09 at 10 09 14 AM
Screen Shot 2022-08-09 at 10 10 39 AM


src/app/cube/new-home/learning-object-info/collections/collections.component.ts
```diff
@@ -7,9 +8,27 @@ import { Component, OnInit } from '@angular/core';
...
+
+  async ngOnInit(): Promise<void> {
+    //get collections
+    await this.collectionService
+    .getCollections()
+    .then((collections: Collection[]) => {
+      this.collections = collections;
+    })
+    .catch(e => {
+      console.error(e.message);
```
mfranz1, line 26: This won't console in production. We can toast the message here. Something generic like "Error! There was a problem fetching collections. Please try again later"

src/app/shared/components/collection-card/collection-card.component.ts
```diff
@@ -21,7 +21,7 @@ export class CollectionCardComponent implements OnInit {
       this.collection.abvName !== 'plan c'
     ) {
       this.pictureLocation =
-        '../../../assets/images/collections/' +
+        '/assets/images/collections/' +
```
mfranz1, line 24: This is a relative path instead of an absolute path. Just curious was anything updated for typescript?

pmanal1 (approved):
Looks good! Just address Mike's comments

katieblomquist:
It's looking good! A couple of quick things on styling: 

- The spacing within the cards looks a little wonky to me, maybe consider making the top/bottom paddings the same?
- Are the button styles used here from the mocks? They look a bit like our old buttons to me. 
- Are there plans to add padding at the homepage level between components rather than margins in the components? If not, maybe consider adding left, right, and bottom margins. 
- It looks like the card's background color and the component background color was inverted. Did we decide to do a darker gray rather than the lighter one in the mocks?

src/app/cube/new-home/learning-object-info/collections/collections.component.ts
JosueMolinaMorales: +1

JosueMolinaMorales (commented):
Just mike and katies comments

src/app/shared/components/collection-card/collection-card.component.ts
AStoerkel1: i don't know of any updates to typescript, but angular has a sourceRoot set to 'src', and `"assets": "src/assets"` is set in the angular.json. For imports you probably need absolute, but for images angular will accept relative path.

EVogel1999 (commented):
Could we make the cards all equal in width like it does on the mobile view?  Or does that make the logos look weird?

AStoerkel1:
> Could we make the cards all equal in width like it does on the mobile view? Or does that make the logos look weird?

the problem is logo size, I'm getting them through collection service and displaying them with an ngFor. If they were all a standard size it would look great, but since they aren't then the NCYTE logo would be going out of the box. Its becoming a real headache on the collection dropdown as well. They all have to be in the same sized square, and NICE is almost out of the box while the NSA logo is barely visible.

JosueMolinaMorales (approved):

katieblomquist:
Looking good! Maybe consider adding a box shadow to the button? You should be able to copy and paste the styling from figma!

katieblomquist (approved):
Looks good!

PR #1502, opened 2022-10-13
Bug/sc 15034/refactor sso redirecturl

AStoerkel1:
This PR fixes a few bugs

- redirect now works with SSO login by setting the redirect in local storage when navigating to the login page
- window does not refresh on login error
- User object is now passed from SSO and set in UserService
- session is set correctly from SSO

### DIFF
--- a/src/app/components/primary-navbar/primary-navbar.component.ts
+++ b/src/app/components/primary-navbar/primary-navbar.component.ts
@@ -44,9 +44,15 @@ export class PrimaryNavbarComponent implements OnInit {
       .subscribe(
         (event: NavigationEvent) => {
           if(event instanceof NavigationStart) {
-            this.redirectUrl = event.url;
+            if(event.url !== '/auth/login'){
+              this.redirectUrl = event.url;
+            }
           }
         });
+  }
+
+  handleRedirect(){
+    localStorage.setItem('ssoRedirect', (this.redirectUrl));
   }
 
   async ngOnInit(): Promise<void> {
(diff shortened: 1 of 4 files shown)


pzalep1 (approved):

src/app/components/primary-navbar/primary-navbar.component.ts
```diff
@@ -44,11 +44,17 @@ export class PrimaryNavbarComponent implements OnInit {
       .subscribe(
         (event: NavigationEvent) => {
           if(event instanceof NavigationStart) {
-            this.redirectUrl = event.url;
+            if(!event.url.includes('/auth/login')){
```
mfranz1, line 47:
```suggestion
            if(!event.url.includes('/auth/login') || !event.url.includes('/auth/register')){
```

mfranz1 (changes requested):

EVogel1999 (approved):

mfranz1 (approved):

src/app/core/auth.service.ts
```diff
@@ -86,24 +86,21 @@ export class AuthService {
...
-          IdentityId: identityId,
-          Token: token
-        }
-      }
-    };
-    this.setSession(sessionObject);
+  public setSsoSession(cookieString: string) {
+    const token = JSON.parse(cookieString);
+      const user = token.user;
+      const tokens: Tokens = {bearer: token.bearer, openId: token.openId};
```
mfranz1, line 96: is openId being used anymore?

Cwagne17 (approved):

pmanal1 (approved):

src/app/core/auth.service.ts
AStoerkel1: Paige says it's for encryption that isn't really needed anymore but it's a required param for setSession() and will break something if we remove it

JosueMolinaMorales (approved):

PR #1508, opened 2022-10-27
Revert "bug/sc-15321/clark naming error no longer shows for special characters"

AStoerkel1:
Reverts Cyber4All/clark-client#1505


AStoerkel1:
👎🏻

PR #1512, opened 2022-10-27
andy and I actually fixed something AND tested it; not that we haven'…

mfranz1:
…t tested before but we were able to test in prod with a real live breathing url and it worked so yeah, eat dirt


AStoerkel1 (approved):

AStoerkel1 (approved):

pmanal1 (approved):
LGTM

PR #1525, opened 2022-11-04
Modal click registers when hovering over overlay

pmanal1:
Story [14635](https://app.shortcut.com/clarkcan/story/14635/modal-click-registers-when-hovering-over-overlay).

## About the Bug
If the mouseup event occurs outside of the pop-up when the mousedown event happens inside the pop-up, it will close. This is problematic since users may click and drag inside an input field and end up with their mouse outside the pop-up, leading to it closing. See the video attached to the story for details.

## Note about the Fix
This is a potential solution to this problem. However, I recognize that the directive is used everywhere else in the client. It can be dangerous to change something here, but I'm also hesitant to create a completely identical directive for the purpose of this story.

### DIFF
--- a/src/app/shared/directives/activate.directive.ts
+++ b/src/app/shared/directives/activate.directive.ts
@@ -14,7 +14,7 @@ export class ActivateDirective {
     this.el.nativeElement.style.cursor = 'pointer';
   }
 
-  @HostListener('click', ['$event'])
+  @HostListener('mousedown', ['$event'])
   @HostListener('keypress', ['$event'])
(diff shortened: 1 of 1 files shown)


mfranz1 (approved):

EVogel1999 (approved):

EVogel1999:
If we ever have the need to retire this directive we should switch to the [dialog directive](https://material.angular.io/components/dialog/overview) from Google that does the same thing.

AStoerkel1 (approved):

JosueMolinaMorales (approved):

pzalep1 (approved):

PR #1564, opened 2023-01-24
Bug/sc 17688/collection page without logo contains collection

JosueMolinaMorales:
This PR adds an `(error)` statement to the img tag that would load a collections logo. If a logo is not found, the default image is used. This eliminates having to list out all the collections that do not have a logo.

### DIFF
--- a/src/app/cube/collection-details/collection-details.component.html
+++ b/src/app/cube/collection-details/collection-details.component.html
@@ -2,7 +2,7 @@
   <section class="hero">
       <div class="inner">
         <div class="logos" [ngClass]="{'long': collection?.abvName === 'ncyte' || collection?.abvName === 'cae_community' || collection?.abvName === 'nice' || collection?.abvName === 'deterlab' || collection?.abvName === '502_project' || collection?.abvName === 'ncc'}">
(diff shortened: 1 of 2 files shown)


src/app/cube/collection-details/collection-details.component.html
```diff
@@ -2,7 +2,7 @@
   <section class="hero">
       <div class="inner">
         <div class="logos" [ngClass]="{'long': collection?.abvName === 'ncyte' || collection?.abvName === 'cae_community' || collection?.abvName === 'nice' || collection?.abvName === 'deterlab' || collection?.abvName === '502_project' || collection?.abvName === 'ncc'}">
-          <img *ngIf="pictureLocation" src="{{ pictureLocation }}" alt="collection logo">
+          <img *ngIf="pictureLocation" src="{{ pictureLocation }}" alt="collection logo" (error)="resetPictureLocation()">
```
mfranz1, line 5:
```suggestion
          <img *ngIf="pictureLocation; else null" src="{{ pictureLocation }}" alt="collection logo">
```
I don't think there's a way to get this to work based on how pictureLocation is declared in the component. But in angular this is valid syntax for an ngIf

mfranz1 (approved):

pzalep1 (approved):

PR #1566, opened 2023-02-01
Implemented default sort on browse page to be newest - oldest

bereketsega:
This PR includes 
1. When a user enters the browse page without a text search the results are ordered newest to oldest.
PR#1566 1

2. When a user enters the browse page with a text search the results are order with the text search and no sort is applied.
PR#1166 2

3. When a user enters the browse page with a filter (topic, collection, guidelines, etc.) the results are sorted from newest to oldest.
PR#1166 3

4. When a user enters a keyword search and the sort option is not selected by them the results are ordered by the keyword search.
PR#1166 4

### DIFF
--- a/src/app/cube/browse/browse.component.ts
+++ b/src/app/cube/browse/browse.component.ts
@@ -33,8 +33,8 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
     guidelines: [],
     level: [],
     standardOutcomes: [],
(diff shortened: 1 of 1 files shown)


JosueMolinaMorales (approved):
Awesome work

pzalep1 (approved):
Great job!

pmanal1 (approved):
very nice

PR #1581, opened 2023-02-21
Action panel not reloading between parent -> child detail pages

stef359:
-Added an "else" statement, so if the revised learning object is not true it's set to null.
Screenshot below of the child detail page before it got reloaded to show it works.
Previously, the action panel would have "edit revision" instead of "create revision", because it was pointing at the parents detail page.
Screenshot 2023-02-21 at 4 18 35 PM

### DIFF
--- a/src/app/cube/details/details.component.ts
+++ b/src/app/cube/details/details.component.ts
@@ -162,7 +162,7 @@ export class DetailsComponent implements OnInit, OnDestroy {
           owners.push(this.releasedLearningObject.author.username);
 
           this.learningObjectOwners = owners;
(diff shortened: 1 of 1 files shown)


JosueMolinaMorales (changes requested):
On the right side, the view revisions button should not be there. This has to do with the `hasRevision` variable we looked at. I think you may have removed this change. Can you fix this?

stef359:
- I just fixed what Josue addressed in my new commit. 
Screenshot below of the page before it got reloaded to show the change (Above the action panel).
Screenshot 2023-02-22 at 10 29 59 AM

JosueMolinaMorales (approved):

PR #1596, opened 2023-04-05
fixed currPage=NaN

bereketsega:
The admin dashboard page query param currPage=NaN is resolved. But we have an issue.

Currently the query parameter won't update, but it works.

https://user-images.githubusercontent.com/72763770/230166278-9621af1a-40b5-4b76-8640-9520ea1f0dbc.mov

But if we have
```diff
-     this.query.currPage = pageNum;
+    this.query = {
+            currPage: pageNum;
+    }
```
in line [147](https://github.com/Cyber4All/clark-client/commit/b66748b9bf74d6435dc3369b2a2ad41aa6808fc3#diff-2fd4bd2324265dd5e8d85292421d4c9d9f6107cf48f47106b768e7e06c6aab28L147), It would update the query param but It won't scroll up.

https://user-images.githubusercontent.com/72763770/230166417-d0c86b4e-3da2-4f43-94b1-ecc39ca4ec59.mov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clark",
-  "version": "5.0.33",
+  "version": "5.0.34",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "clark",
-      "version": "5.0.33",
+      "version": "5.0.34",
       "license": "MIT",
       "dependencies": {
         "@angular/animations": "^13.1.2",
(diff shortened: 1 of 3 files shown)


JosueMolinaMorales (approved):
The reason the page number is not changing is because when the page is changed in `changePage` it is just updating the page number and not the query param but in `set query()` we are updating the page and its query params. If you want to you can also fix this bug, if not you can create a story for it.

Cwagne17 (approved):

mfranz1 (approved):

PR #1607, opened 2023-05-12
Implement an organization dropdown (only in registration, not in edit-profile) 

stef359:
Screenshot 2023-05-12 at 3 22 31 PM
Screenshot 2023-05-12 at 3 23 19 PM
I replaced the text field for the organization in the registration page with a drop down menu that includes all the organizations. 
*I could not replace the text field in edit-profile with the drop down because of an authentication bug. I will be creating a shortcut story on it soon.

### DIFF
--- a/src/environments/environment.ts
+++ b/src/environments/environment.ts
@@ -17,6 +17,7 @@ export const environment = {
   cognitoRegion: 'us-east-1',
   cognitoIdentityPoolId: 'us-east-1:08c3533f-4e0b-4014-9bfe-12a347cb6272',
   cognitoAdminIdentityPoolId: 'us-east-1:eed740a1-fea5-474d-8f57-0036b1871693',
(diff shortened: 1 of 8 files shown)


src/environments/environment.ts
```diff
@@ -17,6 +17,7 @@ export const environment = {
   cognitoRegion: 'us-east-1',
   cognitoIdentityPoolId: 'us-east-1:08c3533f-4e0b-4014-9bfe-12a347cb6272',
   cognitoAdminIdentityPoolId: 'us-east-1:eed740a1-fea5-474d-8f57-0036b1871693',
+  cardOrganizationUrl: 'https://api-gateway.caeresource.directory/organizations?type=&verified=verified&mine=&sort='
```
mfranz1, line 20: This also needs to be added to `environments.prod.ts`

mfranz1 (changes requested):
Just the one comment. But also, any reason why this branch is being merged into the angular 15 branch? This should be based on main

PR #1616, opened 2023-05-24
Implement an organization dropdown (only in registration)

stef359:
*I replaced the text field for the organization in the registration page with a drop down menu that includes all the organizations. 
*I could not replace the text field in edit-profile with the drop down because of an authorization bug.
*This is the third PR I created for this story because of previous issues so everything is from one commit.
Screenshot 2023-05-24 at 10 30 29 AM
Screenshot 2023-05-24 at 10 31 19 AM


src/app/auth/register/register.component.ts
```diff
@@ -340,6 +366,26 @@ export class RegisterComponent implements OnInit, OnDestroy{
     }
   }
 
+  selectOrg(org?: Organization) {
```
bereketsega, line 369: It would be great if we have comment over this function.

bereketsega (commented):
Looks good. Just one suggestion.

mfranz1 (changes requested):
You will also need to add `cardOrganizationUrl` to the `enviornonment.staging.ts`

Cwagne17 (approved):

mfranz1 (approved):

PR #1619, opened 2023-05-30
Bug/sc 23445/make clark logo clickable

shanioffe:
This pull request addresses the following story https://app.shortcut.com/clarkcan/story/23445/make-clark-logo-clickable and makes the CLARK logo open up the homepage in a new tab when clicked.

*This is a redo of a previous pull request that was branched off of the wrong branch.

### DIFF
--- a/src/app/components/primary-navbar/primary-navbar.component.html
+++ b/src/app/components/primary-navbar/primary-navbar.component.html
@@ -1,10 +1,12 @@
 <ng-template #desktop>
     <div class="desktop-navbar" *ngIf="showNav">
         <div class="nav-wrapper">
-            <div class="logo" [routerLink]="['/home']">
-                <span class="logo-image"></span>
-                <span id="clark-logo" class="logo-link">CLARK</span>
-            </div>
+            <a [routerLink]="['/home']" target="_blank" style="text-decoration: none;">    
+                <div class="logo">
+                    <span class="logo-image"></span>
+                    <span id="clark-logo" class="logo-link">CLARK</span>
+                </div>
+            </a>
             <div class="mid-bar">
                 <div class="search-bar">
                     <clark-search></clark-search>
--- a/src/app/core/guideline.service.ts
+++ b/src/app/core/guideline.service.ts
@@ -70,7 +70,7 @@ export class GuidelineService {
     return {
       text: filter.text || filter.filterText,
       year: filter.year !== '' ? filter.year : undefined,
-      levels: filter.levels !== [] ? filter.levels : undefined,
+      levels: filter.levels.length > 0 ? filter.levels : undefined,
       page: filter.page !== '' ? filter.page : undefined,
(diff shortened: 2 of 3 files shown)


src/app/core/guideline.service.ts
```diff
@@ -70,7 +70,7 @@ export class GuidelineService {
     return {
       text: filter.text || filter.filterText,
       year: filter.year !== '' ? filter.year : undefined,
-      levels: filter.levels !== [] ? filter.levels : undefined,
+      levels: filter.levels.length > 0 ? filter.levels : undefined,
```
mfranz1, line 73: Just curious, why is this a part of your PR?

JosueMolinaMorales (approved):

Cwagne17 (approved):

src/app/core/guideline.service.ts
shanioffe: This line was throwing an error that prevented me from running production

mfranz1 (approved):

src/app/components/primary-navbar/primary-navbar.component.html
```diff
@@ -1,10 +1,12 @@
 <ng-template #desktop>
     <div class="desktop-navbar" *ngIf="showNav">
         <div class="nav-wrapper">
-            <div class="logo" [routerLink]="['/home']">
-                <span class="logo-image"></span>
-                <span id="clark-logo" class="logo-link">CLARK</span>
-            </div>
+            <a [routerLink]="['/home']" target="_blank" style="text-decoration: none;">
```
katieblomquist, line 4: Just curious as to why we're opening the link in a new tab. The traditional user flow is to open in the same tab, and I'm just curious as to the reason for the deviation.

katieblomquist, line 4: I'd also suggest moving the in-line styling to the style sheet, but that's a personal preference.

katieblomquist (changes requested):
Looking good! Just a couple of nitpicks.

pmanal1 (approved):
Looks great. Just address Katie's comments

src/app/components/primary-navbar/primary-navbar.component.html
shanioffe: If you take a glance at the story, it asks to make the logo open up the homepage in a new tab. Not sure the reason for the deviation but those were the directions

mfranz1: @katieblomquist good catch! The target should be removed for the home button. Users should remain within CLARK when clicking the logo

shanioffe: Gotchya, I'll get that story revised too so there's no confusion

shanioffe: Good tip! I'll do that

PR #1642, opened 2023-07-24
Bug/sc 23904/remove plan c contribute button

shanioffe:
This PR fixes a bug where when a user visits the Plan C collection, any collections they may visit afterwards also show the contribute button.

Before:
https://github.com/Cyber4All/clark-client/assets/109975145/9bafab00-c766-4a2e-8915-d213aaacb272

After:
https://github.com/Cyber4All/clark-client/assets/109975145/256cb8d7-eb87-491e-9a20-72e9b9cb60d6


src/app/cube/collection-details/collection-details.component.ts
mfranz1, line 1: The story is to remove the button. We can disable is terminally by replacing lines 57 - 60 with just `this.showContribute = false;`. The button should not show at all. We can preserve the HTML code though if we chose to do something like this for a collection in the future

mfranz1 (changes requested):

bereketsega (approved):
Just Mike's comment

JosueMolinaMorales (commented):
Just mike's comment

PR #1645, opened 2023-08-04
Change first name appearing as last name

stef359:
When a person registers in clark with a first name that has two names in it, the first name variable of the user would always end up being the first part of their first name and the second part of their first name ends up with the last name. I will attach pictures to show it.
Initial Problem:
image
image
'Mary Grace' is the first name but 'Grace' ends up in the last name after the edit submission.
After:
Screenshot 2023-08-04 at 12 06 39 AM
Screenshot 2023-08-04 at 12 07 22 AM

This is supposed to fix that problem, starting from registration all the way to the edit profile component. The first name with two names would stay in the first name variable with no change.

### DIFF
--- a/src/app/core/user.service.ts
+++ b/src/app/core/user.service.ts
@@ -242,6 +242,28 @@ export class UserService {
     return deleteValue;
   }
 
+  combineName(firstname: string, lastname: string){
+    let firstnameArray = firstname.split(' ');
+    let lastnameArray = lastname.split(' ');
+
+    if(firstnameArray.length > 1){
+      let newfirstName = '';
+      for(let i = 0; i < firstnameArray.length - 1; i++){
+      newfirstName += firstnameArray[i] + '#';
+      }
+      newfirstName += firstnameArray[firstnameArray.length - 1];
+      firstname = newfirstName;
+  } if(lastnameArray.length > 1){
+    let newlastName = '';
+    for(let i = 0; i < lastnameArray.length - 1; i++){
+    newlastName += lastnameArray[i] + '#';
+    }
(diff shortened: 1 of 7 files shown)


src/app/core/user.service.ts
```diff
@@ -242,6 +242,28 @@ export class UserService {
     return deleteValue;
   }
 
+  combineName(firstname: string, lastname: string){
+    let firstnameArray = firstname.split(' ');
+    let lastnameArray = lastname.split(' ');
```
bereketsega, line 247: The names might consist trailing spaces and we can remove them with `trim()` so that we don't have empty strings in our array.
```suggestion
    let firstnameArray = firstname.trim().split(' ');
    let lastnameArray = lastname.trim().split(' ');
```

src/app/core/user.service.ts
```diff
@@ -242,6 +242,28 @@ export class UserService {
...
+
+    if(firstnameArray.length > 1){
+      let newfirstName = '';
+      for(let i = 0; i < firstnameArray.length - 1; i++){
+      newfirstName += firstnameArray[i] + '#';
+      }
+      newfirstName += firstnameArray[firstnameArray.length - 1];
+      firstname = newfirstName;
+  } if(lastnameArray.length > 1){
```
bereketsega, line 256:
```suggestion
  } 
  if(lastnameArray.length > 1){
```

bereketsega (changes requested):
Looks good. Just few suggestions.

shanioffe (approved):
Looks good! Just BK's comments

mfranz1 (approved):

bereketsega (approved):

PR #1670, opened 2023-10-30
Add browse button to secondary nav bar

stef359:
Added a Browse button on the nav bar.
Screenshot 2023-10-30 at 2 29 31 PM


katieblomquist (changes requested):
My only comment is that Browse is the only CTA we have in the list, and should either be first or last. I would ask a higher power than myself as to which is preferred. Otherwise, this looks good!

stef359:
The changes made to the header component file are from a previous task. I don't know why the changes are showing up on this branch even thought they were committed and pushed a while ago.

Cwagne17 (approved):

shanioffe (approved):

bereketsega (approved):

PR #1686, opened 2024-02-05
Chore/sc 28570/remove lighthouse from clard stack

shanioffe:
This PR removes the lighthouse dependency from `clark-client`. I was not able to find any lighthouse dependencies in another repository, including `cards-client`, so this should be the only PR needed for this story. 

Note: `package-lock` has some references to a package called lighthouse-logger, we can ignore that because it is being used as a dependency by other packages.


mfranz1 (approved):
Good stuff!

bereketsega (approved):

stef359 (approved):

Cwagne17 (approved):

PR #1702, opened 2024-03-11
rating and metrics module resolved

Cwagne17:
## What this PR does / why we need it

The changes in the PR reviews the changes to the clark-client specifically for the ratings and metrics module. It removes any routes that were previously used, cleans up the services for ratings and metrics and looks to move them to the /src/app/core directory.

### Added

- getLearningObjectStats to the metrics service
- getUserMetrics to the metrics service
- Added LearningObjectRatings and Rating types to rating service
- Added a return type to getLearningObjectRatings

### Changed

- Moved GET_LEARNING_OBJECT_STATS to the legacy learning object routes since it has not been implemented in clark-service yet
- Updated the parameter for GET_COLLECTION_METRICS to match backend
- Moved BloomsDistribution interface to the metrics module to support getLearningObjectStats
- Updated getCollectionMetrics handling of HTTP error
- Changed getLearningObjectRatings to use more descriptive mapping variables
- Changed type for ratings in the details component to avoid conflict with HTTP return type

### Removed

- GET_RATINGS from the rating routes since it isn't used in the client

## Acceptance Criteria Met

- [ ] All review conversations resolved

**Special notes for your reviewer**:


katieblomquist:
Could you add a description for work done?

src/app/cube/core/cube-core.module.ts
```diff
@@ -1,6 +1,6 @@
 import { NgModule, Optional, SkipSelf } from '@angular/core';
 import { CommonModule } from '@angular/common';
-import { UsageStatsService } from './usage-stats/usage-stats.service';
+import { MetricService } from 'app/core/metric-module/metric.service';
```
mfranz1, line 3:
```suggestion
```

src/app/cube/core/cube-core.module.ts
```diff
@@ -10,7 +10,7 @@ import { UsageStatsService } from './usage-stats/usage-stats.service';
 @NgModule({
   imports: [CommonModule],
   exports: [],
-  providers: [UsageStatsService]
+  providers: [MetricService]
```
mfranz1, line 13:
```suggestion
```
This can be deleted because of the way angular works. You'll notice in each service there is a `@Injectable{providedIn: root}` This allows us to use a service where ever in the client. We don't need to provide the service

src/app/cube/details/details.component.ts
mfranz1, line 1: Nice. Did you test this by chance?

mfranz1 (changes requested):
Looks good. Just remove the service from the provider pls

src/app/cube/home/splash/splash.component.ts
```diff
@@ -11,12 +11,12 @@ export class SplashComponent implements OnInit {
...
   constructor(
-    private usageStatsService: UsageStatsService,
+    private metricService: MetricService,
     public googleTagService: GoogleTagService
     ) { }
 
   ngOnInit(): void {
-    this.usageStatsService.getLearningObjectStats().then(stats => {
+    this.metricService.getLearningObjectStats().then(stats => {
       this.numReleasedObjects = Math.floor(stats.released / 10) * 10;
```
shanioffe, line 20: Just a question on this math. Why don't we just set numReleasedObjects = stats.released? Why do we divide by 10 and then multiply by 10?

JosueMolinaMorales: We do this so that we can display a round number in the home component. 

This is how its used:
image

For example if we have 1595 objects we divide by 10 to get 159.5, floor it to get 159 then multiply by 10 to get 1590. This way we can just add a `+` to the end of it.

JosueMolinaMorales (approved):

mfranz1 (approved):

PR #1705, opened 2024-03-14
update files for revisions

JosueMolinaMorales:


shanioffe (approved):

src/app/onion/learning-object-builder/builder-store.service.ts
```diff
@@ -245,7 +245,7 @@ export class BuilderStore {
     const retrieve = this._isRevision && revisionId !== undefined && username ? async () => {
       // eslint-disable-next-line eqeqeq
       if (revisionId == 0) {
-        revisionId = await this.learningObjectService.createRevision(username, id);
+        revisionId = await this.learningObjectService.createRevision(username);
```
mfranz1, line 248: This is correct? username is an optional parameter to this service function but to the `learning-object.service.ts` it expects cuid. This param should be id? Also, could you verify what id is being passed. I could help test if you would like to pair

mfranz1: Check id meaning please confirm it is cuid and not just an object id. And could the function block comment be updated

Cwagne17: +1

PR #1721, opened 2024-04-25
Mfranz1/clark reports fix

mfranz1:
# Description
Two issues occurred:
1. The lambda went ideal due to lack of use.
2. It seems at some point authorization was droped and/or added to this route resulting in the 401 being thrown

Fixes:
* Lambda was restarted
* Event bridge rule to hit the service
* Added authorization headers through client api request

### DIFF
--- a/src/app/core/collection.service.ts
+++ b/src/app/core/collection.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { HttpClient, HttpErrorResponse } from '@angular/common/http';
+import { HttpClient, HttpErrorResponse, HttpHeaders } from '@angular/common/http';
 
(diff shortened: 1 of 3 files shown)


pzalep1 (approved):

src/app/core/collection.service.ts
```diff
@@ -242,7 +243,8 @@ export class CollectionService {
           {
             email,
             name
-          }
+          },
+          { headers: this.headers, withCredentials: true }
```
JosueMolinaMorales, line 247: Where are headers initialized? Will it automatically put the bearer token?

JosueMolinaMorales (commented):
1 question

src/app/core/collection.service.ts
pzalep1: this.headers does include that in the services on the client

mfranz1: Up on line 22 we declare our Headers

JosueMolinaMorales (approved):
lgtm

PR #1723, opened 2024-04-26
Changes for updating (patch) learning outcome mappings story

Greek:
https://github.com/Cyber4All/clark-client/assets/17275723/dd03ca18-d5f5-4f60-bf2e-cea42cbb3025

https://app.shortcut.com/clarkcan/story/28559/implement-updating-the-mappings-that-a-guideline-affects


Greek:
by the way this won't work for you probably, I still need to patch up the LO topics implementation as well which the relevancy builder uses.

Cwagne17 (approved):

PR #1732, opened 2024-05-11
[sc-28554] implement removing files from a los bundle

Greek:
Updates the implementation for toggling file bundle states on the client, see https://app.shortcut.com/clarkcan/story/28554/implement-removing-files-from-a-los-bundle-into-clark-service 

Example:
https://github.com/Cyber4All/clark-client/assets/17275723/2dd0a253-1984-4347-aefc-dac32aaf8015

### DIFF
--- a/src/app/onion/core/learning-object.service.ts
+++ b/src/app/onion/core/learning-object.service.ts
@@ -51,19 +51,17 @@ export class LearningObjectService {
   /**
    * Calls LO service to update the packageable status of toggled files
    *
-   * @param username The currently logged in user
    * @param learningObjectID The current learning object's ID
    * @param fileIDs An array of file IDs that need to be updated
    * @param state The new packageable property to update to
    * @returns A promise
    */
   toggleBundle(
-    username: string,
     learningObjectId: string,
     fileIDs: string[],
     state: boolean
   ) {
-    const route = BUNDLING_ROUTES.TOGGLE_BUNDLE_FILE({ username, learningObjectId });
+    const route = BUNDLING_ROUTES.TOGGLE_BUNDLE_FILE({ learningObjectId });
 
     return this.http
       .patch(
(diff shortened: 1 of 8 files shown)


Greek:
i don't think i can test this properly till we get learning-object module working again. for now i'll mark the story as blocked and try it again later

src/entity/learning-object/learning-object.ts
```diff
@@ -838,7 +838,8 @@ export namespace LearningObject {
 
   export namespace Material {
     export interface File {
-      id: string;
+      _id: string;
+      id: string; // Keeping for backward compat, may wanna remove?
```
shanioffe, line 842: Hmm, unless someone says definitely to remove it, you could mark `id` as deprecated, or you could change its name to something like "source" like the way [learning-outcome model](https://github.com/Cyber4All/clark-service/blob/main/src/shared/types/learning-outcome.type.ts) is implemented.

src/app/onion/core/learning-object.service.ts
```diff
@@ -51,26 +51,24 @@ export class LearningObjectService {
...
-    const route = BUNDLING_ROUTES.TOGGLE_BUNDLE_FILE({ username, learningObjectId });
+    const route = BUNDLING_ROUTES.TOGGLE_BUNDLE_FILE({ learningObjectId });
 
     return this.http
       .patch(
         route,
         {
           fileIDs: fileIDs,
-          state: state
+          packagable: state
```
mfranz1, line 71: thank you

src/entity/learning-object/learning-object.ts
mfranz1: I believe it is safe to remove `id`. Please just double check where the File interface is being implemented/referenced throughout the client.

curlyLasagna (approved):

JosueMolinaMorales (approved):

src/entity/learning-object/learning-object.ts
Greek: done. the client builds, i believe this change should be okay to deploy =)

mfranz1 (approved):

shanioffe (approved):

PR #1750, opened 2024-06-19
feature/sc-29042/implement-getting-a-learning-objects-outcomes

johngonz0:
Client changes for getting a learning objects outcomes


shanioffe (approved):

mfranz1 (approved):
John and myself (for all thats worth) searched far and wide and there's no reference to this route throughout the client. Outcomes are a resourceURI on an object so we use the route returned from the service. We will leave this here just in case it is needed in the future for something

Greek (approved):

JosueMolinaMorales (approved):

PR #1765, opened 2024-06-28
feat: [sc-29548] update route for uploading file metadata

Greek:
**Note:** Cannot test this yet. File upload functionality seems to be broken because client is missing some data from the backend that is required to "upload" the file to S3, assuming that S3 even would work in this case.

### DIFF
--- a/src/app/onion/learning-object-builder/builder-store.service.ts
+++ b/src/app/onion/learning-object-builder/builder-store.service.ts
@@ -694,7 +694,6 @@ export class BuilderStore {
     await this.learningObjectService
       .addFileMeta({
         files,
-        username: this.learningObject.author.username,
         objectId: this.learningObject._id
       })
       .then(() => {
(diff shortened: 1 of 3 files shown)


src/app/onion/learning-object-builder/builder-store.service.ts
mfranz1, line 1: technically the parameters are flipped, I guess that doesn't matter too much to typescript?

stef359 (approved):

curlyLasagna (approved):

src/app/onion/learning-object-builder/builder-store.service.ts
Greek: should be fine cuz it's an object

JosueMolinaMorales (approved):

pzalep1 (approved):
Looks good!

johngonz0 (approved):

mfranz1 (approved):

PR #1775, opened 2024-07-15
Fix Login

pzalep1:
This PR adds an explicit setting of the cookie. This follows the same pattern as the competency client.


johngonz0 (approved):
yay!

mfranz1 (approved):

PR #1778, opened 2024-07-16
bug: [sc-31788] fix incorrect child LOs being removed

Greek:
See [shortcut story](https://app.shortcut.com/clarkcan/story/31788/bug-removing-children-from-learning-object-in-builder) for information on the bug.

The `#setChildren` function in the `BuilderStore` class used a list of children that was stored within that instance of `BuilderStore`. Therefore when removing a child from the list, the LO inside the builder store instance was never updated, and old data was used to update the LO. The solution was to fetch the LO again in the builder store *before* setting the children.

also yes this was a one line fix. :godmode:


pzalep1 (approved):
Chef's kiss. 
image

johngonz0 (approved):
yay'

PR #1782, opened 2024-07-18
bug: [sc-32272] add back `id` field to learning object

Greek:
Some routes in clark-service return the ID field as `id` instead of the standard `_id`. Changing these would take a bit, so I added the `id` field to the frontend as a "quick fix".

This fixes multiple issues in some routes where the `id` field does not exist when calling `__id` or `_id`. 

As far as the Shortcut story goes, the problem has been addressed including possibly many other issues.

### DIFF
--- a/src/entity/learning-object/learning-object.ts
+++ b/src/entity/learning-object/learning-object.ts
@@ -27,6 +27,10 @@ export class LearningObject {
     }
   }
 
+  get id(): string {
+    return this.__id;
+  }
+
   private _cuid?: string;
 
   /**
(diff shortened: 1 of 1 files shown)


src/entity/learning-object/learning-object.ts
```diff
@@ -638,8 +642,10 @@ export class LearningObject {
    * @memberof LearningObject
    */
   private copyObject(object: Partial<LearningObject>): void {
-    if (object._id) {
-      this._id = object._id;
+    if (object.id) {
+      this.__id = object.id;
+    } else if (object._id) {
+      this.__id = object._id;
```
mfranz1, line 648: The FullLearningObject _id is a mongo id of the objects-index collection which cannot be used to make api requests. John has a PR up that address the id issue.

mfranz1 (changes requested):
This cannot be merged due to incompatible ids

src/entity/learning-object/learning-object.ts
mfranz1: FullLearningObject is the return from clark-service*^

PR #1789, opened 2024-07-25
bug/sc-32353/editing-lo-from-admin-dashboard-searches

johngonz0:
## What this PR fixes

When editing a learning object from the admin dashboard, the builder route was hit with learning object Id. I corrected it to use the cuid instead.

I also noticed another builder re-route that was still using learning object id for editors, so I changed that as well.

https://github.com/user-attachments/assets/f01d3f60-db20-45c2-81cb-52b814808d66

Screenshot 2024-07-25 at 1 48 28 PM


pzalep1 (approved):

Greek (approved):
YAY

stef359 (approved):

curlyLasagna (approved):

mfranz1 (approved):

PR #1792, opened 2024-07-27
Bug/sc 32315/bundle learning object route returns 400

curlyLasagna:
## Description

Rewrite download bundle since auth headers are required to be passed to the backend

## Fixes 
- Filter out library items that doesn't contain a learning object property
- User library items will not have their id's dropped

### DIFF
--- a/src/app/core/library-module/library.service.ts
+++ b/src/app/core/library-module/library.service.ts
@@ -39,8 +39,10 @@ export class LibraryService {
     // this.headers.append('Content-Type', 'application/json');
   }
 
-  async getLibrary(opts?: { learningObjectCuid?: string, version?: number,
-                  page?: number, limit?: number }): Promise<{ cartItems: LearningObject[], lastPage: number }> {
+  async getLibrary(opts?: {
+    learningObjectCuid?: string, version?: number,
+    page?: number, limit?: number
+  }): Promise<{ cartItems: LearningObject[], lastPage: number }> {
     this.updateUser();
     if (!this.user) {
       return Promise.reject('User is undefined');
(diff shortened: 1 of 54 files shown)


src/app/core/library-module/library.service.ts
```diff
@@ -63,7 +65,14 @@ export class LibraryService {
...
       .then((val: any) => {
-        this.libraryItems = val.userLibraryItems.map(object => new LearningObject(object.learningObject));
+        this.libraryItems = val.userLibraryItems
+          .filter(object => object && object.learningObject) // Filter out objects without learningObject
+          .map(object => {
+            if (object.learningObject) {
+              object.learningObject.id = object.learningObject._id;
+            }
+            return new LearningObject(object.learningObject);
+          });
```
mfranz1, line 75: Based on the staging data, this object would not be undefined. Lets pull staging data to ensure all values are being returned. Otherwise, there's a bigger issue we need to address.

image

src/app/core/library-module/library.service.ts
```diff
@@ -124,56 +133,86 @@ export class LibraryService {
...
+      // REQUIRED: Set the download attribute to the name of the file
+      link.download = this.getBundleName(contentDisposition);
+      document.body.appendChild(link);
+      // Trigger the download
+      link.click();
+      // Remove the element from the DOM
+      document.body.removeChild(link);
+      // Revoke the object URL to prevent memory leaks
+      window.URL.revokeObjectURL(link.href);
+    });
```
mfranz1, line 215: Absolutely beautiful like 13/10 top programmers right here who ever did this 🙃

mfranz1 (approved):

mfranz1:
dismissed my approval until staging data is tested with implementation

pzalep1 (approved):

mfranz1 (approved):

src/app/core/library-module/library.service.ts
curlyLasagna: Out of scope. Totally different bug that I'll create another story for

stef359 (approved):

mfranz1 (approved):

PR #1793, opened 2024-07-30
fix: [sc-32405] add more allowed status changes

Greek:
Some status changes that are allowed don't show up in the client. This isn't an exhaustive change and may be more statuses that should be allowed in the client.

See https://app.shortcut.com/clarkcan/story/32405/some-status-changes-are-not-allowed.

### DIFF
--- a/src/app/onion/learning-object-builder/components/editor-action-panel/change-status-modal/change-status-modal.component.ts
+++ b/src/app/onion/learning-object-builder/components/editor-action-panel/change-status-modal/change-status-modal.component.ts
@@ -82,6 +82,10 @@ export class ChangeStatusModalComponent implements OnInit {
         break;
       case LearningObject.Status.PROOFING:
         this.statuses = [LearningObject.Status.RELEASED, LearningObject.Status.REJECTED];
+        break;
+      case LearningObject.Status.RELEASED:
+        this.statuses = [LearningObject.Status.PROOFING];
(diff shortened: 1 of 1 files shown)


src/app/onion/learning-object-builder/components/editor-action-panel/change-status-modal/change-status-modal.component.ts
```diff
@@ -77,13 +77,15 @@ export class ChangeStatusModalComponent implements OnInit {
...
       case LearningObject.Status.ACCEPTED_MAJOR:
-        this.statuses = [LearningObject.Status.WAITING];
+        this.statuses = [LearningObject.Status.WAITING, LearningObject.Status.PROOFING];
         break;
       case LearningObject.Status.PROOFING:
         this.statuses = [LearningObject.Status.RELEASED, LearningObject.Status.REJECTED];
+        break;
+      case LearningObject.Status.RELEASED:
+        this.statuses = [LearningObject.Status.PROOFING];
+        break;
```
pzalep1, line 88: This is actually handled by a separate function in the admin dashboard so it is not needed here.

pzalep1 (commented):
Looks good. Just the one comment.

mfranz1 (approved):

pzalep1 (approved):

PR #1798, opened 2024-07-31
Bug/sc 32418/metrics throwing 404 error on details page

shanioffe:
## Description

Implements learning object metrics in client (this is not being used because we get the URI in the full learning object builder in clark-service, but I implemented anyway to minimize future tech debt).

Screenshot 2024-07-31 at 1 25 03 PM

Before (no numbers)

Screenshot 2024-07-31 at 1 24 50 PM

After with numbers!

### DIFF
--- a/src/app/core/metric-module/metric.routes.ts
+++ b/src/app/core/metric-module/metric.routes.ts
@@ -14,7 +14,7 @@ export const METRIC_ROUTES = {
    * @returns metrics for one learning object or all released learning objects
    */
   GET_LEARNING_OBJECT_METRICS(cuid?: string) {
-    return `${environment.apiURL}/learning-objects/metrics?cuid=${cuid}`;
+    return `${environment.apiURL}/learning-objects/metrics?${querystring.stringify({cuid: cuid})}`;
   },
 
   /**
--- a/src/app/core/metric-module/metric.routes.ts
+++ b/src/app/core/metric-module/metric.routes.ts
@@ -13,8 +13,8 @@ export const METRIC_ROUTES = {
    * Gets learning object metrics
    * @returns metrics for one learning object or all released learning objects
    */
-  GET_LEARNING_OBJECT_METRICS() {
-    return `${environment.apiURL}/learning-objects/metrics`;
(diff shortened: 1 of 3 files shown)


curlyLasagna (approved):
LGTM 🚀

src/app/core/metric-module/metric.routes.ts
```diff
@@ -13,8 +13,8 @@ export const METRIC_ROUTES = {
    * Gets learning object metrics
    * @returns metrics for one learning object or all released learning objects
    */
-  GET_LEARNING_OBJECT_METRICS() {
-    return `${environment.apiURL}/learning-objects/metrics`;
+  GET_LEARNING_OBJECT_METRICS(cuid?: string) {
+    return `${environment.apiURL}/learning-objects/metrics?cuid=${cuid}`;
```
Greek, line 17: use querystring to pass `cuid` instead of appending it to the string. this is done so if cuid is empty, we won't pass an empty string which could result in the backend returning nothing

```suggestion
    return `${environment.apiURL}/learning-objects/metrics?${querystring.stringify({cuid: cuid})}`;
```

Greek (changes requested):

src/app/core/metric-module/metric.routes.ts
shanioffe: Okay, thanks

pzalep1 (approved):
Beautiful.

mfranz1 (approved):

src/app/core/metric-module/metric.routes.ts
```diff
@@ -14,7 +14,7 @@ export const METRIC_ROUTES = {
    * @returns metrics for one learning object or all released learning objects
    */
   GET_LEARNING_OBJECT_METRICS(cuid?: string) {
-    return `${environment.apiURL}/learning-objects/metrics?cuid=${cuid}`;
+    return `${environment.apiURL}/learning-objects/metrics?${querystring.stringify({cuid: cuid})}`;
```
curlyLasagna, line 17: image

PR #1805, opened 2024-08-02
Fix client sending learning object id instead of cuid

curlyLasagna:
When editing a learning object's status to released, it goes to map and tag but the function that routes towards map and tag passes learning object ID instead of CUID, so client is confusion

### DIFF
--- a/src/app/onion/learning-object-builder/components/editor-action-panel/change-status-modal/change-status-modal.component.ts
+++ b/src/app/onion/learning-object-builder/components/editor-action-panel/change-status-modal/change-status-modal.component.ts
@@ -91,7 +91,6 @@ export class ChangeStatusModalComponent implements OnInit {
   private moveToMapAndTag() {
     // Set a small timeout before navigating so that the change in status applies
     setTimeout(() => {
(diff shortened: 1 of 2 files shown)


src/app/onion/learning-object-builder/components/editor-action-panel/change-status-modal/change-status-modal.component.ts
```diff
@@ -91,7 +91,8 @@ export class ChangeStatusModalComponent implements OnInit {
   private moveToMapAndTag() {
     // Set a small timeout before navigating so that the change in status applies
     setTimeout(() => {
-      this.router.navigate(['onion', 'relevancy-builder', this.learningObject.id, 'outcomes']);
+      console.log('cuid:',this.learningObject.cuid);
```
curlyLasagna, line 94: calling myself out before anyone does

🔥🔥🔥🔥

johngonz0 (approved):

pzalep1 (approved):

mfranz1 (approved):

PR #1807, opened 2024-08-06
fix: [sc-32512] clicking save to library gives no final feedback

Greek:
This PR fixes issues where no feedback is given to the user when saving a learning object to their library

The fixes in particular involved the removal of a nested try-catch block, which prevented the toaster from opening, and improved error handling to only handle unsuccessful requests.

https://github.com/user-attachments/assets/ed49c222-faaf-40f4-bc13-9e26058778a4

### DIFF
--- a/src/app/core/library-module/library.service.ts
+++ b/src/app/core/library-module/library.service.ts
@@ -49,10 +49,10 @@ export class LibraryService {
     }
 
     const query = new URLSearchParams({
-      page: opts.page ? opts.page.toString() : '1',
-      limit: opts.limit ? opts.limit.toString() : '10',
-      cuid: opts.learningObjectCuid ? opts.learningObjectCuid : '',
-      version: opts.version ? opts.version.toString() : '0'
+      page: opts?.page ? opts.page.toString() : '1',
+      limit: opts?.limit ? opts.limit.toString() : '10',
+      cuid: opts?.learningObjectCuid ? opts.learningObjectCuid : '',
+      version: opts?.version ? opts.version.toString() : '0'
     });
 
     return await this.http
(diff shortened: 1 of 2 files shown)


src/app/cube/details/components/action-panel/action-panel.component.ts
```diff
@@ -142,26 +143,30 @@ export class ActionPanelComponent implements OnInit, OnDestroy {
...
         }
       }
-    } catch (error) {
-      this.toaster.error('Error!', 'There was an error adding to your library');
     }
-    this.libraryService.getLibrary();
+
+    this.toaster.success('Successfully Added!', 'Learning Object added to your library');
+    this.saved = true;
+    this.animateSaves();
```
pzalep1, line 167: If it doesn't break something else (and knowing this action panel component it might) I believe this can get moved up into the try where you add to the library so the success, saved state change, and animations are all together.

mfranz1 (approved):

src/app/core/library-module/library.service.ts
```diff
@@ -49,10 +49,10 @@ export class LibraryService {
     }
 
     const query = new URLSearchParams({
-      page: opts.page ? opts.page.toString() : '1',
-      limit: opts.limit ? opts.limit.toString() : '10',
-      cuid: opts.learningObjectCuid ? opts.learningObjectCuid : '',
-      version: opts.version ? opts.version.toString() : '0'
+      page: opts?.page ? opts.page.toString() : '1',
```
curlyLasagna, line 52: nice! less noise on the console

curlyLasagna (approved):

pzalep1 (approved):
Looks good!

stef359 (approved):

PR #1812, opened 2024-08-07
update user route to conform to backend

curlyLasagna:
Update client to use backend route

This resolves the issue of user not being able to save their information


pzalep1 (approved):

johngonz0 (approved):
nice

stef359 (approved):

PR #1828, opened 2024-08-20
fix: [sc-32644] Curator - Filtering in curator dashboard always returns all released learning objects

Greek:
The admin dashboard would apply filters to its query three times and the incorrect data would get returned. This PR fixes that issue, and the filter data should only be applied once to the query on initial page load.

See accompanying service PR: https://github.com/Cyber4All/clark-service/pull/201
Story details: https://app.shortcut.com/clarkcan/story/32644

https://github.com/user-attachments/assets/1f2a3bae-75e8-4e99-8c3a-4fe3a561e897

### DIFF
--- a/src/app/admin/components/filter-search/filter-search.component.ts
+++ b/src/app/admin/components/filter-search/filter-search.component.ts
@@ -78,28 +78,42 @@ export class FilterSearchComponent implements OnInit {
     //check for params in the query and add them to the filter dropdown bars
     const qParams = this.route.parent.snapshot.queryParams;
 
+    /**
+     * TODO: refactor this to send a complete search query
+     * at the first request instead of sending a request
+     * every time we emit an event on the toggle filter functions.
+     *
+     * this COULD be a cause of the client sending bad data back
+     * to the user, so just be aware.
+     */
+
     //if there are topics in the query, toggle them in the filter dropdown
     if (qParams.topics && qParams.topics.length > 0) {
       const topics = [];
         for (const topic of qParams.topics) {
           topics.push({ name: '', _id: topic });
         }
+
         this.toggleTopicFilter(topics[0]);
     }
 
-    //if there are statuses in the query add them too the filter dropdowns
+    // if there are statuses in the query add them to the filter dropdowns
     if (qParams.status && qParams.status.length > 0) {
-        const statuses = [];
+      const statuses = [];
+
+      if (Array.isArray(qParams.status) === false) {
+        statuses.push(qParams.status);
+        this.toggleStatusFilter(statuses);
+
+      } else {
         for (const status of qParams.status) {
           statuses.push(status);
         }
-
-        console.log(statuses);
-
         this.toggleStatusFilter(statuses);
+      }
     }
-    //if there is a collection selected in the query, toggle it
-    //there will only ever be a single collection selected at a time
+
+    // Set the collection if it exists, otherwise default back to the params.
     if (qParams.collection) {
       this.toggleCollectionFilter(qParams.collection);
     } else {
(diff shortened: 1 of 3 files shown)


pzalep1 (approved):

stef359 (approved):

johngonz0 (approved):

src/app/admin/components/filter-search/filter-search.component.ts
```diff
@@ -77,34 +77,51 @@ export class FilterSearchComponent implements OnInit {
...
-    //if there are statuses in the query add them too the filter dropdowns
-    if (qParams.status) {
-      //multiple statuses in query
-      if (Array.isArray(qParams.status)) {
+
+    // if there are statuses in the query add them to the filter dropdowns
+    if (qParams.status && qParams.status.length > 0) {
+      const statuses = [];
+
+      if (Array.isArray(qParams.status) === false) {
```
Cwagne17, line 108: Nit: 

```suggestion
      # If the query param `status` is not an array
      if (!Array.isArray(qParams.status)) {
```

src/app/admin/components/filter-search/filter-search.component.ts
```diff
@@ -77,34 +77,51 @@ export class FilterSearchComponent implements OnInit {
...
+        this.toggleTopicFilter([{ name: '', _id: qParams.topics }]);
       }
     }
-    //if there are statuses in the query add them too the filter dropdowns
-    if (qParams.status) {
-      //multiple statuses in query
-      if (Array.isArray(qParams.status)) {
+
+    // if there are statuses in the query add them to the filter dropdowns
+    if (qParams.status && qParams.status.length > 0) {
```
Cwagne17, line 105: How come we have to check is qParams.status is an array on ln108 if we are checking `.length` on this line?

If status is a string then it is checking characters, if it array then its checking elements. I don't know if we need to check length in this case.

src/app/admin/components/filter-search/filter-search.component.ts
```diff
@@ -77,34 +77,51 @@ export class FilterSearchComponent implements OnInit {
...
+     * this COULD be a cause of the client sending bad data back
+     * to the user, so just be aware.
+     */
+
     //if there are topics in the query, toggle them in the filter dropdown
-    if (qParams.topics) {
-      //more than one topic
+    if (qParams.topics && qParams.topics.length > 0) {
+      const topics = [];
       if (Array.isArray(qParams.topics)) {
```
Cwagne17, line 93: Instead of having to do this type checking can we enforce the type sent to the component?

Cwagne17 (commented):
Just a few questions.

src/app/admin/components/filter-search/filter-search.component.ts
Greek: removed this for converting to a map instead

src/app/admin/components/filter-search/filter-search.component.ts
Greek: removed

src/app/admin/components/filter-search/filter-search.component.ts
```diff
@@ -71,40 +72,45 @@ export class FilterSearchComponent implements OnInit {
...
+      (s) => !['rejected', 'unreleased'].includes(s.toLowerCase()),
     );
     this.relevancyStart = new Date();
 
     //check for params in the query and add them to the filter dropdown bars
-    const qParams = this.route.parent.snapshot.queryParams;
+    const qParams = this.route.parent.snapshot.queryParamMap;
+
+    /**
+     * TODO: refactor this to send a complete search query
```
Cwagne17, line 83: Can we make a story for this, so that we can track this need.

Cwagne17 (commented):
Looks good brief comment.

PR #1829, opened 2024-08-23
fix: [sc-32712] Unable to add child as an author admin

Greek:
### I wasn't able to reproduce this bug, this change might or might not fix anything.

When trying to add a new child to a *new* learning object, the search will break because it couldn't find the username of the author.

### DIFF
--- a/src/app/onion/learning-object-builder/components/scaffold/add-child/add-child.component.ts
+++ b/src/app/onion/learning-object-builder/components/scaffold/add-child/add-child.component.ts
@@ -52,7 +52,8 @@ export class AddChildComponent implements OnInit, OnDestroy {
    */
   async getLearningObjects(filters?: any, query?: string): Promise<LearningObject[]> {
     this.loading = true;
-    const draftObjects = await this.learningObjectService.getDraftLearningObjects(this.child.author.username, { ...filters, text: query })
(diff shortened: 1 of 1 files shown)


src/app/onion/learning-object-builder/components/scaffold/add-child/add-child.component.ts
```diff
@@ -52,7 +52,8 @@ export class AddChildComponent implements OnInit, OnDestroy {
    */
   async getLearningObjects(filters?: any, query?: string): Promise<LearningObject[]> {
     this.loading = true;
-    const draftObjects = await this.learningObjectService.getDraftLearningObjects(this.child.author.username, { ...filters, text: query })
+    const draftObjects = await this.learningObjectService.getDraftLearningObjects(this.child.author.username || this.auth.username,
```
Cwagne17, line 55: I don't think we want to use the auth.username as the alternative because it looks like we are specifically getting draft objects for the learning object's children. Can we guarantee that `this.auth.username` is the owner of the child object? Because I though a lot of the implementations where this.auth is actually the user logged in.

Cwagne17 (commented):
Question about using the username from this.auth.

stef359 (approved):

PR #1837, opened 2024-08-30
Feature/sc 32766/feature module route updates for clark refactor

Romerico234:
The route UPDATE_COLLECTION_FEATURED_OBJECTS is not referenced anywhere. All other routes are good.


src/app/core/feature-module/featured.routes.ts
Cwagne17, line 1: Can we combine `GET_COLLECTION_FEATURED_OBJECTS` and `GET_COLLECTION_FEATURED_OBJECTS_WITH_LIMIT` one just uses an optional query. These could be combined into one function.

Cwagne17 (changes requested):

src/app/core/feature-module/featured.routes.ts
Cwagne17: Can we also update the names in the `featued.service.ts` to match the naming of the routes so it is more identifiable?

Cwagne17: Can we also add types to the function parameters in the `featured.service.ts`?

pzalep1: ^

pzalep1 (commented):
Looks good after the comments from Chris are addressed.

Romerico234:
Combined GET_COLLECTION_FEATURED_OBJECTS and GET_COLLECTION_FEATURED_OBJECTS_WITH_LIMIT. Updated function parameter name for GET_COLLECTION_FEATURED_OBJECTS in featured.service.ts. Added types to function parameters.

src/app/core/feature-module/featured.routes.ts
```diff
@@ -10,26 +10,22 @@ export const FEATURED_ROUTES = {
...
+  GET_COLLECTION_FEATURED_OBJECTS(
+    collectionAbvName: string,
+    limit?: number,
+  ) {
+    if (typeof limit !== 'undefined') {
+      return `${environment.apiURL}/featured/learning-objects?collection=${encodeURIComponent(
+        collectionAbvName,
+      )}&limit=${encodeURIComponent(limit)}`;
+    }
     return `${environment.apiURL}/featured/learning-objects?collection=${encodeURIComponent(collectionAbvName)}`;
```
Cwagne17, line 27: I'd recommend doing this type of check. Using `if(<variable>)` will check if the variable exists. Also instead of creating two unique strings you can append the additional query to the first string.

```suggestion
    let url = `${environment.apiURL}/featured/learning-objects?collection=${encodeURIComponent(collectionAbvName)}`;
    if (limit) {
          url += `&limit=${encodeURIComponent(limit)}`;
    }
    return url;
```

Cwagne17 (commented):
One minor change. Great work.

Cwagne17 (approved):

pzalep1 (approved):

PR #1843, opened 2024-09-04
feat: [sc-32770] User Module route updates for clark-refactor

Greek:
Story details: https://app.shortcut.com/clarkcan/story/32770


src/app/auth/register/components/recaptcha/recaptcha-validator.service.ts
```diff
@@ -14,7 +14,7 @@ export class RecaptchaValidator {
 
     validateToken(token: string) {
```
Cwagne17, line 15: Nit: Can this be moved to the user service?

src/app/core/learning-object-module/editor.service.ts
```diff
@@ -21,7 +21,7 @@ export class EditorService {
     authorUsername: string,
     status: string,
   ): Observable<string> {
-    const route = LEGACY_USER_ROUTES.UPDATE_MY_LEARNING_OBJECT(authorUsername, id);
+    const route = LEARNING_OBJECT_ROUTES.UPDATE_LEARNING_OBJECT_STATUS(id);
```
Cwagne17, line 24: Can this be moved to the learning object service?

src/app/core/user-module/user.service.ts
```diff
@@ -95,6 +75,16 @@ export class UserService {
         );
   }
 
+  getUsersLearningObjects(username: string): Promise<LearningObject[]> {
+    return this.http
+      .get(USER_ROUTES.GET_USER(username), { withCredentials: true })
```
Cwagne17, line 80: Nit: Can we update one of the names either the function or the route name. because one looks like we are getting all learning objects for a user and the other looks like we are getting the user's information.

Cwagne17: I just looked at the route in the USER_ROUTES. that is the route to get the user's document in the users collection not the route to get learning objects for a user.

src/app/core/user-module/user.service.ts
```diff
@@ -95,6 +75,16 @@ export class UserService {
...
+      .pipe(catchError(this.handleError))
+      .toPromise()
+      .then((val: any) => {
+        return val.map((l) => new LearningObject(l));
```
Cwagne17, line 84: This will return a mostly empty LearningObject because it is trying to map a User to a LearningObject.

src/app/onion/core/learning-object.service.ts
```diff
@@ -143,26 +138,6 @@ export class LearningObjectService {
...
       });
   }
-  /**
-   * Fetches Learning Object by ID (full)
-   *
-   * @param {string} learningObjectID
-   * @returns {Promise<LearningObject>}
-   * @memberof LearningObjectService
-   */
-  getLearningObject(learningObjectId: string): Promise<LearningObject> {
```
Cwagne17, line 153: Was this not being used?

src/app/shared/components/user-card/user-card.component.ts
```diff
@@ -25,11 +25,12 @@ export class UserCardComponent implements OnInit, OnChanges {
   }
 
   ngOnChanges() {
+    // XXX: When is this called??
     this.fetchLearningObjects();
   }
 
   async fetchLearningObjects() {
-    this.objects = await this.learningObjectService.getUsersLearningObjects(this.user.username);
+    this.objects = await this.userService.getUsersLearningObjects(this.user.username);
```
Cwagne17, line 33: This is the route that should be used for this.

This is from clark-service, it should be in the /learning-objects/search/search.service.ts

```typescript
router.get(
            "/users/:username/learning-objects",
            this.getUsersLearningObjects,
        );
```

Cwagne17 (changes requested):

src/app/core/user-module/user.routes.ts
```diff
@@ -45,14 +45,8 @@ export const USER_ROUTES = {
...
   },
-  /**
-   * Request to retrieve a user's profile
-   * @method GET
-   * @auth required
-   * @param username the username of the user
-   * @returns the user's profile
-   */
-  GET_USER_PROFILE(username: string) {
-    return `${environment.apiURL}/users/${encodeURIComponent(username)}/profile`;
```
johngonz0, line 56: why was this removed?

johngonz0 (commented):
just a question

src/app/core/user-module/user.routes.ts
Greek: it's not used anywhere and i haven't seen any mention of it on the analysis docs

src/app/onion/core/learning-object.service.ts
Greek: mentioned this in the standup: this route was used when we used to fetch learning objects by their IDs, not their CUIDs. so the logic that use this function and the function itself were removed as the need for querying by ID isn't really needed anymore.

src/app/core/user-module/user.service.ts
Greek: changed to use appropriate route =)

src/app/core/learning-object-module/editor.service.ts
Greek: i removed the entire route because it wasn't used here, but we should probably remove the entirety of `EditorService` too because it's not referenced anywhere.

i'll keep `EditorService` for now though

src/app/auth/register/components/recaptcha/recaptcha-validator.service.ts
Greek: it was moved to the user service but i'm not able to test it because the Register page doesn't work, which is its only use.

stef359 (approved):

johngonz0 (approved):

src/app/shared/components/user-card/user-card.component.ts
```diff
@@ -25,11 +25,12 @@ export class UserCardComponent implements OnInit, OnChanges {
   }
 
   ngOnChanges() {
+    // XXX: When is this called??
     this.fetchLearningObjects();
   }
```
pzalep1, line 30: ngOnChanges is a lifecycle hook. Here is an explanation: https://v17.angular.io/guide/lifecycle-hooks

pzalep1 (commented):
Left a comment for the lifecycle hook explanation.

src/app/core/learning-object-module/learning-object/learning-object.routes.ts
```diff
@@ -122,6 +137,28 @@ export const LEARNING_OBJECT_ROUTES = {
...
+          mappingsId: string,
+      ) {
+          return `${environment.apiURL}/users/${encodeURIComponent(
+              username,
+          )}/learning-objects/${encodeURIComponent(
+              learningObjectId,
+          )}/outcomes/${encodeURIComponent(outcomeId)}/mappings/${encodeURIComponent(
+              mappingsId,
+          )}`;
+      },
```
Cwagne17, line 161: Are these the paths that are used? I thought most of the routes we got rid of `/users/${username}...` Are these from standard guidelines?

src/app/core/user-module/user.service.ts
```diff
@@ -146,6 +137,24 @@ export class UserService {
       .toPromise();
   }
 
+  /**
+   * Validate a user's captcha token
+   *
+   * @param {string} token the token to verify
+   * @returns an object with the result if fail, or null if true.
+   */
+  validateToken(token: string) {
```
Cwagne17, line 146: Can we call this validate captcha instead of validate token because validate token is a separate route that we have in auth

src/app/core/user-module/user.routes.ts
```diff
@@ -28,6 +28,15 @@ export const USER_ROUTES = {
   GET_USER(user: string) {
     return `${environment.apiURL}/users/${encodeURIComponent(user)}`;
   },
+  /**
+   * Request to retrieve a user's learning objects
+   * @method GET
+   * @param user the username or userId of the user
+   * @returns the learning objects
+   */
+  GET_USER_LEARNING_OBJECTS(user: string) {
```
Cwagne17, line 37: This route should be in the learning-object-module/search

src/app/core/learning-object-module/learning-object/learning-object.routes.ts
```diff
@@ -122,6 +137,28 @@ export const LEARNING_OBJECT_ROUTES = {
...
+            learningObjectId,
+        )}/outcomes/${encodeURIComponent(outcomeId)}/mappings`;
+    },
+    DELETE_MAPPING(
+          username: string,
+          learningObjectId: string,
+          outcomeId: string,
+          mappingsId: string,
+      ) {
+          return `${environment.apiURL}/users/${encodeURIComponent(
```
Cwagne17, line 154: I'm pretty sure this all got updated and should be in the learning-object-module/outcomes module. There aren't any routes called /mappings in the gateway or backend

Screenshot 2024-09-09 at 9 37 20 AM

Cwagne17 (changes requested):
Some routes need to be moved to the module according to how the gateway and clark-service have them organized.

dsoto18 (approved):

stef359 (approved):

johngonz0 (approved):

dsoto18 (approved):

PR #1845, opened 2024-09-05
Feature/sc 32767/report module route updates for clark refactor

Cwagne17:
this story updates the collections service to move it's reports route into the reports.service.ts.


johngonz0 (approved):
fire

pzalep1 (approved):

PR #1846, opened 2024-09-05
Feature/sc 32768/library module route updates for clark refactor

Cwagne17:
This story fixes the library service routes and broken logic that didn't allow for pagination and simplified some roundabout legacy logic of getting library _id by using the learningObject that exists on the document etc. etc..

### DIFF
--- a/src/app/cube/library/library.component.ts
+++ b/src/app/cube/library/library.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit, OnDestroy, HostListener, ChangeDetectorRef, ViewChild, ElementRef } from '@angular/core';
-import { LibraryService } from 'app/core/library-module/library.service';
+import { LibraryItem, LibraryService } from 'app/core/library-module/library.service';
 import { LearningObject } from 'entity/learning-object/learning-object';
 import { ToastrOvenService } from 'app/shared/modules/toaster/notification.service';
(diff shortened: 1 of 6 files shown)


src/app/cube/library/library.component.ts
```diff
@@ -124,17 +124,20 @@ export class LibraryComponent implements OnInit, OnDestroy {
...
+
+      this.libraryItems.map(async (libraryItem: LibraryItem) => {
+        const ratings = await this.getRatings(libraryItem.learningObject);
         if (ratings) {
           libraryItem['avgRating'] = ratings.avgValue;
         }
       });
       this.loading = false;
     } catch (e) {
+      console.log(e);
```
johngonz0, line 140: is this an intentional console.log

johngonz0 (approved):
looks good just one question

src/app/cube/library/library.component.ts
pzalep1: The library does not handle errors.

pzalep1 (approved):

stef359 (approved):

src/app/cube/library/library.component.ts
Cwagne17: Yes because otherwise we wouldn't know what the issue was in the client.

PR #1848, opened 2024-09-05
File module route updates

stef359:
Just relocated some functions using file routes to file.service.ts and updated their references in the components they are used in.

Story: https://app.shortcut.com/clarkcan/story/32778/learningobject-file-module-route-updates-for-clark-refactor


johngonz0 (approved):
nice

src/app/core/learning-object-module/file/file.service.ts
```diff
@@ -53,6 +59,198 @@ export class FileService {
...
+  addFileMeta({
+    objectId,
+    files
+  }: {
+      objectId: string;
+      files: FileUploadMeta[];
+    }): Promise<string[]> {
+      const route = FILE_ROUTES.UPLOAD_FILE_META(objectId);
+      return this.handleFileMetaRequests(files, route);
+    }
```
Cwagne17, line 150: Nit: If nothing else uses handleFileMetaRequests, can we just combine the functions so it is more clear to what is happening?

src/app/core/learning-object-module/file/file.service.ts
```diff
@@ -53,6 +59,198 @@ export class FileService {
...
+      return fileIds;
+    }
+
+  /**
+   * Handles downloading a file by opening the stream url in a new window
+   *
+   * @param {LearningObject.Material.File} file[The file to be downloaded]
+   * @memberof UploadComponent
+   */
+    async handleFileDownload(file: LearningObject.Material.File) {
```
Cwagne17, line 247: What component uses this? It looks like there aren't any references.

Cwagne17 (commented):
Two questions.

src/app/core/learning-object-module/file/file.service.ts
pzalep1: +1

pzalep1 (commented):
Just Chris's comment.

src/app/core/learning-object-module/file/file.service.ts
stef359: Good catch, I ended up moving the function back to its original component file because it was directly used in the corresponding HTML file.

Cwagne17: If the function is using a file.routes.ts it should be moved to the file.service.ts but we just need to import the FileService into the component's constructor.

Cwagne17 (commented):
Response to the initial question.

src/app/core/learning-object-module/file/file.service.ts
stef359: Okay it's done ✅

Cwagne17 (approved):

pzalep1 (approved):

dsoto18 (approved):

PR #1858, opened 2024-09-13
Fixing a bug in determining whether we need a changelog

Cwagne17:
We never were able to set needs changelog to true since it defaults to false.


pzalep1 (approved):
Gorgeous.

stef359 (approved):

PR #1859, opened 2024-09-13
Feature/sc 32769/standard guidelines module route updates

Romerico234:
Removed unused routes and added types to some function parameters.

### DIFF
--- a/src/app/core/standard-guidelines-module/standard-guidelines.service.ts
+++ b/src/app/core/standard-guidelines-module/standard-guidelines.service.ts
@@ -2,20 +2,30 @@ import { Injectable } from '@angular/core';
 import { HttpClient, HttpErrorResponse } from '@angular/common/http';
 import * as querystring from 'querystring';
 import { throwError } from 'rxjs';
-import { retry, catchError, map } from 'rxjs/operators';
+import { catchError, map } from 'rxjs/operators';
 import { STANDARD_GUIDELINES_ROUTES } from './standard-guidelines.routes';
 import { FrameworkDocument } from '../../../entity/standard-guidelines/Framework';
 import { SearchItemDocument } from '../../../entity/standard-guidelines/search-index';
 
+interface GuidelinesFilter {
+  text?: string;
+  filterText?: string;
+  year?: string;
+  levels?: string;
+  page?: number;
+  limit?: number;
+  type?: string;
+  frameworks?: string | string[];
+}
+
 @Injectable({
-  providedIn: 'root'
+  providedIn: 'root',
 })
 export class GuidelineService {
-
-  constructor(public http: HttpClient) { }
+  constructor(public http: HttpClient) {}
 
   getGuidelines(
-    filter?
+    filter?: GuidelinesFilter,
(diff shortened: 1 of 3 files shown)


joaquinVill (approved):

src/app/core/standard-guidelines-module/standard-guidelines.service.ts
```diff
@@ -15,7 +15,7 @@ export class GuidelineService {
   constructor(public http: HttpClient) { }
 
   getGuidelines(
-    filter?
+    filter?: any
```
Cwagne17, line 18: Are we able to add a more specific type other then any?

src/app/core/standard-guidelines-module/standard-guidelines.service.ts
```diff
@@ -47,7 +46,7 @@ export class GuidelineService {
       ).toPromise();
   }
 
-  private formatFilter(filter) {
+  private formatFilter(filter: any) {
```
Cwagne17, line 49: Same question here.

You can refer to standard-guidelines to check out what it is expecting.

dsoto18:
Apart from Chris's comment looks good to me.

src/app/core/standard-guidelines-module/standard-guidelines.service.ts
pzalep1: There is probably a type for this over in standard guidelines: https://github.com/Cyber4All/standard-guidelines-service

Romerico234:
Created an interface to define the possible inputs in the filters parameter--named GuidelinesFilter.

src/app/core/standard-guidelines-module/standard-guidelines.service.ts
```diff
@@ -2,52 +2,62 @@ import { Injectable } from '@angular/core';
...
+        total: number;
+        results: SearchItemDocument[];
+      }>(STANDARD_GUIDELINES_ROUTES.SEARCH_GUIDELINES())
+      .pipe(catchError(this.handleError))
+      .toPromise();
+    return res;
   }
 
   // CLARK-SERVICE-FIX: SUPPORT SEARCH QUERY
   async getFrameworks(query?: any): Promise<FrameworkDocument[]> {
```
Greek, line 46: optional: remove `any` here and replace it with the appropriate type. i think it might be the same `GuidelinesFilter` ([see here](https://github.com/Cyber4All/standard-guidelines-service/blob/main/src/modules/search-module/types/SearchQueryArgs.ts#L4)) though i'm not too sure. not too big of a deal if not you can't find the type fort his.

Greek (approved):

src/app/core/standard-guidelines-module/standard-guidelines.service.ts
pzalep1: +1

pzalep1 (commented):
Looks good! After the comment from Andreas is addressed.

stef359 (approved):
Andreas' comment then it should be good 👍

Cwagne17 (approved):

PR #1863, opened 2024-09-16
fix: [sc-32930] Add support for username on admin dashboard

Greek:
This PR adds support for querying a user's learning objects with a username query. The admin "users" page utilizes this functionality.

Story details: https://app.shortcut.com/clarkcan/story/32930

### DIFF
--- a/src/app/admin/pages/learning-objects/learning-objects.component.ts
+++ b/src/app/admin/pages/learning-objects/learning-objects.component.ts
@@ -22,10 +22,11 @@ import { UserService } from 'app/core/user-module/user.service';
   selector: 'clark-learning-objects',
   templateUrl: './learning-objects.component.html',
   styleUrls: ['./learning-objects.component.scss'],
-  providers: [PublicLearningObjectService]
+  providers: [PublicLearningObjectService],
 })
 export class LearningObjectsComponent
-  implements OnInit, OnDestroy, AfterViewInit {
+  implements OnInit, OnDestroy, AfterViewInit
+{
   @ViewChild('list') listElement: ElementRef<HTMLElement>;
   @ViewChild('headers') headersElement: ElementRef<HTMLElement>;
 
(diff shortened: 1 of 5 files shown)


Greek (commented):
this is all linting

johngonz0 (approved):

src/app/admin/pages/learning-objects/learning-objects.component.ts
```diff
@@ -211,29 +217,68 @@ export class LearningObjectsComponent
...
-        })
-        .finally(() => {
-          this.loading = false;
-        });
+      if (this.query.username && this.query.username.length > 0) {
+        await this.searchService
+          .getUsersLearningObjects(this.query.username, {
+            ...this.query,
+            // too lazy to lowercase allStatuses values
+            status: allStatuses.map((v) => v.toLowerCase()),
```
Greek, line 240: i didn't wanna use draftsOnly here because then we'd have to add draft LOs onto the list of released LOs and that would break pagination, so I just append all the statuses instead..

there might be a way to append draft LOs with released LOs without breaking pagination i briefly looked at.. i'd be happy to re-visit that solution again if needed

pzalep1: Appending the statuses is fine here from what I can tell.

pzalep1 (approved):

src/app/admin/pages/learning-objects/learning-objects.component.ts
```diff
@@ -211,29 +217,68 @@ export class LearningObjectsComponent
    * @memberof LearningObjectsComponent
    */
   async getLearningObjects() {
+    const allStatuses = [
+      'UNRELEASED',
```
Cwagne17, line 221: I think there is an enum we can reference to prevent hardcoding this.

src/app/admin/pages/learning-objects/learning-objects.component.ts
Cwagne17: If we are doing toLowerCase we should just make the array as lowercase statuses, but also I think there is enum somewhere that has it stored that way already.

Cwagne17 (commented):
Looks good, just had the suggestion of using an enum for status types.

stef359 (approved):

PR #1864, opened 2024-09-16
route fixes

joaquinVill:
change-authorship.service.ts was not being used, it held an empty class. The route was going to authorship.service.ts instead. I updated route so that the change-authorship.service.ts was being used instead. Should I delete the authorship.service.ts file since it isnt being used anymore?

### DIFF
--- a/src/app/admin/core/authorship.service.ts
+++ b/src/app/admin/core/authorship.service.ts
@@ -1,49 +1,8 @@
 import { Injectable } from '@angular/core';
-import { catchError } from 'rxjs/operators';
-import { HttpClient, HttpErrorResponse, HttpHeaders } from '@angular/common/http';
-import { throwError } from 'rxjs';
-import { User } from '@entity';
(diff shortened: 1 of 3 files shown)


src/app/admin/core/authorship.service.ts
```diff
@@ -1,49 +1,8 @@
...
-import { catchError } from 'rxjs/operators';
-import { HttpClient, HttpErrorResponse, HttpHeaders } from '@angular/common/http';
-import { throwError } from 'rxjs';
-import { User } from '@entity';
-import { CHANGE_AUTHORSHIP_ROUTES } from '../../core/learning-object-module/change-authorship/change-authorship.routes';
 
 @Injectable({
   providedIn: 'root'
 })
 export class AuthorshipService {
```
Cwagne17, line 6: We can delete this whole file if it doesn't have any functions in it anymore.

Cwagne17 (commented):
Looks good, just need to remove the old authorship.service.ts file and lint the project

Greek (approved):

Cwagne17 (approved):

dsoto18 (approved):

PR #1865, opened 2024-09-18
Fix broken preview URL being generated

Cwagne17:
Hopefully this fixes the CORS issue...

Kinda hard to replicate CORS locally. This fix at-least will authenticate the request to preview and return the blob then the client will render the blob in a new tab.


pzalep1 (approved):
Welp ... lettuce see if it works in staging too.

stef359 (approved):

PR #1870, opened 2024-09-20
update register and interceptor

Cwagne17:
I updated the interceptor to get the token if it doesn't have it already. This was an issue I found when registering it wouldn't have the token because the provider was already intialized.

Also it updates the body sent to the register route so that the name is split into firstname and lastname.

### DIFF
--- a/src/app/core/auth-module/auth.service.ts
+++ b/src/app/core/auth-module/auth.service.ts
@@ -388,7 +388,14 @@ export class AuthService {
    *
    * @param user the user's data from the registration component
    */
-  async register(user: any): Promise<User> {
+  async register(user: {
+    username: string,
+    firstname: string,
+    lastname: string,
+    email: string,
+    organization: string,
+    password: string,
+  }): Promise<User> {
     try {
       const data = await this.encryptionService.encryptRSA(user);
       const response = await this.http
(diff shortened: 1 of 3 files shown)


src/app/core/auth-module/auth.service.ts
```diff
@@ -388,7 +388,14 @@ export class AuthService {
    *
    * @param user the user's data from the registration component
    */
-  async register(user: any): Promise<User> {
+  async register(user: {
```
dsoto18, line 391: Looks good, there isn't a User DTO we could use? Just a thought

johngonz0 (approved):
nice

src/app/core/auth-module/auth.service.ts
Greek: ^ there's a million places in the client that need proper typing too, would be a good story to complete later on =)

Greek (approved):

src/app/core/auth-module/auth.service.ts
pzalep1: Believe it or not, but there used to be an npm package that provided types for everything. It was removed in 2019.

pzalep1 (approved):

src/app/core/auth-module/auth.service.ts
Cwagne17: @dsoto18 There is a type in the backend, but like Andreas said, as you'll come to learn there are many refactors needed on the frontend for proper typing.

I do agree we should get to this though

PR #1874, opened 2024-10-04
bug: [sc-33807] No children returned in learning object builder

Greek:
The text parameter was set as "undefined" which returns... nothing because there's no LOs with "undefined" in them. Now, if the text truly is undefined, we return an empty string.

This fixes the issue where no objects are returned when there is no input text.

https://app.shortcut.com/clarkcan/story/33807/no-children-returned-in-learning-object-builder


joaquinVill (approved):

johngonz0 (approved):

stef359 (approved):

pzalep1 (approved):
Beautiful.

Cwagne17 (approved):

dsoto18 (approved):

PR #1876, opened 2024-10-04
Bug/sc 33809/an author can add other learning objects

joaquinVill:
This fixes an author being able to add learning objects as children that are not owned by that author. This also fixes the search mechanism that was not returning any learning objects. I also consolidated the search.service.ts files into one file instead of 2.

Story: https://app.shortcut.com/clarkcan/story/33809/an-author-can-add-other-learning-objects-as-children

Video:
- learning Object "Wow this is awesome2" is a released learning object authored by Paige (who I am signed in as)
- the rest of the learning objects that appear in the search are draft learning object authored by Paige

https://github.com/user-attachments/assets/72b53464-dd79-442c-bf6a-f90603cb018e


src/app/onion/learning-object-builder/components/scaffold/add-child/add-child.component.ts
```diff
@@ -62,20 +62,8 @@ export class AddChildComponent implements OnInit, OnDestroy {
...
-      .then((response: { learningObjects: LearningObject[], total: number }) => {
-        let { learningObjects } = response;
-        const indx = this.lengths.indexOf(this.child.length);
-        const childrenLengths = this.lengths.slice(0, indx);
-        learningObjects = learningObjects.filter(
-          child => (!this.currentChildren.includes(child.id) && childrenLengths.includes(child.length))
-        );
-        return learningObjects;
-      });
-
```
pzalep1, line 76: I would see if you could adjust the query to just include the author username. An author can add released learning objects as children of unreleased as long as they are the authors and are of a shorter length.

Cwagne17 (changes requested):

johngonz0 (approved):
This looks really good!

src/app/core/learning-object-module/search/search.service.ts
```diff
@@ -75,4 +78,8 @@ export class SearchService {
       return throwError(error);
     }
   }
+
+  registerChange() {
+    this.needsChange$.next();
+  }
```
pzalep1, line 84: Where is this called in the service?

joaquinVill: Only reference is in the add-evaluator.component.ts file. Should I move this to that file since its not used anywhere else?
Screenshot 2024-10-10 at 2 16 51 PM

dsoto18 (approved):

stef359 (approved):

Cwagne17 (approved):

src/app/core/learning-object-module/search/search.service.ts
pzalep1: I would see who is subscribed to this and then make the call.

pzalep1 (approved):

PR #1887, opened 2024-12-12
fix dist with images in root of build

Cwagne17:


src/app/onion/core/suggestion.service.ts
```diff
@@ -1,9 +1,9 @@
 import { Observable, Subject, throwError } from 'rxjs';
-import { retry, catchError } from 'rxjs/operators';
+import { catchError } from 'rxjs/operators';
 import { Injectable } from '@angular/core';
 import { HttpClient, HttpHeaders, HttpErrorResponse } from '@angular/common/http';
 import { environment } from '@env/environment';
-import * as querystring from 'querystring';
+import { stringify } from 'app/core/shared/stringify';
```
pzalep1, line 7: Double check the scss imports for this component are still good since you moved the location.

PR #1896, opened 2024-12-17
update search bar < 3 char

johngonz0:
This PR updates the searching functionality on the client. For any searches < 3 characters, a 3 second error message pops up, and the search is not performed. All other search functionality functions as intended.

See video of testing

https://github.com/user-attachments/assets/61704f44-2776-458d-b501-b5d7087ae8ab

### DIFF
--- a/src/app/components/search/search.component.html
+++ b/src/app/components/search/search.component.html
@@ -22,10 +22,7 @@
       </div>
     </form>
     <!-- Alert message for invalid input -->
-    <div
-      id="search-alert"
-      style="color: red; font-weight: bold; display: none; margin-top: 5px;"
-    >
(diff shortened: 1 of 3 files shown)


src/app/components/search/search.component.html
```diff
@@ -21,5 +21,13 @@
         </button>
       </div>
     </form>
+    <!-- Alert message for invalid input -->
+    <div
+      id="search-alert"
+      style="color: red; font-weight: bold; display: none; margin-top: 5px;"
```
pzalep1, line 27: This should get put in the scss file

Cwagne17 (approved):
I thought < 3 was a ❤️  lol

PR #1900, opened 2024-12-18
fix: [sc-35541] 'Users undefined' when viewing learning object

Greek:
Action buttons were returning an error when the user isn't authenticated. This change makes an early return instead when the user isn't authenticated.

Story details: https://app.shortcut.com/clarkcan/story/35541


Cwagne17 (approved):

pzalep1 (approved):

pzalep1:
I'm going to merge this with this mornings release.

PR #1935, opened 2024-12-31
Topics selected from the navbar not loading after clear-filters is clicked 

stef359:
Added JS functionality to reload the page whenever a topic is selected from the dropdown, ensuring the page updates with the newly applied topic.

**_Problem_**
If a user interacted with the browse component by applying filters, clearing filters, or searching text—the nav bar's topic dropdown would became unresponsive after. As a result, selecting a topic from the nav bar dropdown would not apply the filter unless the user manually reloaded the page. 

Before:

https://github.com/user-attachments/assets/8c157b75-4488-4392-bf9e-21e75bbae5d8

After:

https://github.com/user-attachments/assets/835ccaaa-8d8a-45dd-976d-e5cf61c3c309

### DIFF
--- a/src/app/core/client-module/navBarDropdown.service.ts
+++ b/src/app/core/client-module/navBarDropdown.service.ts
@@ -50,8 +50,9 @@ export class NavbarDropdownService {
 
     public setTopic(topic: Topic): void {
         this.closeAll();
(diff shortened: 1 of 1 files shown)


pzalep1 (approved):

johngonz0 (approved):
nice

PR #1940, opened 2025-01-07
2 bug fixes: Password reset sending 2 emails and curator dashboard showing all released objects

pzalep1:
# Description 
The submit function was being called by the button and the ngSubmit functionality in the reset password request form. I removed the call on the button to only send the request to the backend once. 

This also resolves the issue with the curator dashboard showing all learning objects instead of just the collection that the curator is a part of.

### DIFF
--- a/src/app/core/learning-object-module/search/search.service.ts
+++ b/src/app/core/learning-object-module/search/search.service.ts
@@ -30,7 +30,6 @@ export class SearchService {
     let route = '';
     if (query) {
       const queryClone = Object.assign({}, query);
-      console.log('called');
       if (
         queryClone.standardOutcomes &&
         queryClone.standardOutcomes.length &&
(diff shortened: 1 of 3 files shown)


dsoto18 (approved):
Nice fix, are we keeping the console log in the search service?

src/app/core/learning-object-module/search/search.service.ts
```diff
@@ -30,6 +30,7 @@ export class SearchService {
     let route = '';
     if (query) {
       const queryClone = Object.assign({}, query);
+      console.log('called');
```
johngonz0, line 33: 🤑

Cwagne17 (approved):

stef359 (approved):

johngonz0 (approved):

dsoto18 (approved):

PR #1962, opened 2025-02-04
fix: [sc-35850] Look into collection statistics issues

Greek:
The NCyTE dashboard stats would return an incorrect count of released learning objects because we were using a total count of learning objects, instead of the count for released learning objects.

See clark-service PR [here](https://github.com/Cyber4All/clark-service/pull/342)
Story details: https://app.shortcut.com/clarkcan/story/35850


dsoto18 (approved):

pzalep1 (approved):
I am going to merge this so it goes in today

PR #1965, opened 2025-02-07
Bug/sc 35726/fix navigation options not sticking to secondary

dsoto18:


dsoto18:
https://github.com/user-attachments/assets/373f9972-9aa9-46f5-8e84-a6b7fb083f7a

johngonz0 (approved):
I have no idea how this works, but this is a great fix!

src/app/shared/modules/contextmenu/context-menu-viewer/context-menu-viewer.component.ts
```diff
@@ -34,10 +37,31 @@ export class ContextMenuViewerComponent {
...
+      this.activateClose();
+    }
+  }
+
+  @HostListener('window:scroll', ['$event'])
+  onScroll(){
+    // close context menu when user scrolls past all the options
+    this.activateClose();
+  }
+
```
johngonz0, line 59: What happens if i scroll up with the dropdown open? Will this close it as well? Just curious... I dont really think it matters that much considering your at the top of the page

joaquinVill (approved):
cool

Romerico234:
I think this was also an issue for the context menu for the user's profile, does this fix that one as well?

dsoto18:
> I think this was also an issue for the context menu for the user's profile, does this fix that one as well?

Just had to make a few changes for this, now it fixes the issue on that and on user dashboard for options on their learning objects. The change was that I had to account for the context menu possibly containing icons with the svg tag, thanks for pointing out @Romerico234

src/app/shared/modules/contextmenu/context-menu-viewer/context-menu-viewer.component.ts
dsoto18: Yes it would close scrolling up too, and yea not too much a concern because they're at the top and the use would have to be scrolled down very slightly to not be at the very top and still be able to open it, but anyways yes it would @johngonz0

Romerico234 (approved):

pzalep1 (approved):

PR #1994, opened 2025-03-12
Feature/sc 36166/implement learning objects by length chart

dsoto18:
The pie chart is implemented on the CyberSkills2Work dashboard displaying the length metrics for the learning objects in their collection. The chart displays the correct data from the new lengthMetrics property returned in the GET /metrics route, and is responsive to give exact values when portions of it are clicked. The pie chart along with the top downloads table are integrated with the clark-ncyte-stats component which appears in both the ncyte dashboard and ncyte collection page. The extra boolean value was added to the stats component so that we can toggle between dashboard view which displays the pie chart and top download metrics, and the collection page view which should not.

### DIFF
--- a/src/app/collection/cyberskills-dashboard/cyberskills-dashboard.component.html
+++ b/src/app/collection/cyberskills-dashboard/cyberskills-dashboard.component.html
@@ -1,7 +1,7 @@
 <div class="container">
     <h1 class="title">CyberSkills2Work Collection Statistics Dashboard</h1>
   
-    <clark-ncyte-stats [collectionName]="'cyberskills2work'" [dashboardView]="true"></clark-ncyte-stats>
+    <clark-dashboard-stats [collectionName]="'cyberskills2work'" [dashboardView]="true"></clark-dashboard-stats>
 
     <div class="downloads-header">
       <h2>All Learning Objects</h2>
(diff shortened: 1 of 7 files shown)


dsoto18:
https://github.com/user-attachments/assets/690b81a7-ff93-40ed-a46f-85eb3d22a4e6

Romerico234 (commented):
I think it would be better to create a separate stats component (`clark-cyber2skillswork-stats`) to avoid coupling it with the NCyTE stats component.

johngonz0 (approved):

Greek (approved):
looks good! i have the concern as Mico though, we should build a more general component for this.

dsoto18:
the ncyte stats component name was changed to just be dashboard stats, which takes an input to conditionally turn on the components for most downloaded learning objects and the learning objects length chart. There already is another stats component that only shows the 'Curriculum By The Numbers' cards, which can be refactored later to hold everything and conditionally turn on and off components so that this can be the only stats component we need. This will be part of a future story that will get done after cyberskills2work @Romerico234 @Greek @pzalep1

Romerico234 (approved):
Looks good, just one nit pick. Can you change `dashboardView` to be something a little more specific like `pieChartView`?

PR #2001, opened 2025-03-18
:sparkles: Add specifications for deprecated query

pzalep1:
# Description 
This PR adds the deprecated query to areas of the client where we are querying for the guidelines. 

# Details
In the builder we want to only return non-deprecated alignments for new learning objects or learning objects that are being updated. 

On the browse page we want to return all alignments since there are learning objects still mapped to the old stuff and people should still be able to search by it.


pzalep1:
> nice, did we mean to change the apiURL in the environment.ts?

Yes, environment.ts is used in the local environment prod is used in production.

PR #2005, opened 2025-03-20
fix: [sc-36240] [Browse] LO Description appears to be truncated/removed incorrectly due to anchor tag

Greek:
There is an issue on the client where descriptions are prematurely cut off because of closing anchor tags, which cause the entirety of the text beyond that tag to be stripped as a result of the regex.

The fix was to edit the regex to match all tags and strip them accordingly.

Story details: https://app.shortcut.com/clarkcan/story/36240

### DIFF
--- a/src/app/cube/shared/learning-object/learning-object.component.ts
+++ b/src/app/cube/shared/learning-object/learning-object.component.ts
@@ -118,7 +118,7 @@ export class LearningObjectListingComponent implements OnInit, OnChanges, OnDest
     // The top regex is used for matching tags such as <br />
     // The bottom regex will catch tags such as </p>
     str = str.replace(/<[0-z\s\'\"=]*[\/]+>/gi, ' ');
-    return str.replace(/<[\/]*[0-z\s\'\"=]+>/gi, ' ');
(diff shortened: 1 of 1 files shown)


dsoto18 (approved):

src/app/cube/shared/learning-object/learning-object.component.ts
```diff
@@ -118,7 +118,7 @@ export class LearningObjectListingComponent implements OnInit, OnChanges, OnDest
     // The top regex is used for matching tags such as <br />
     // The bottom regex will catch tags such as </p>
     str = str.replace(/<[0-z\s\'\"=]*[\/]+>/gi, ' ');
-    return str.replace(/<[\/]*[0-z\s\'\"=]+>/gi, ' ');
+    return str.replace(/<(?!\/?a\s+href="[^"\/\"]*")[^>]+>/gi, ' ');
```
pzalep1, line 121: Let's hold off on this merge this until we talk about if there is more to this. I think this covers one case, but there is a few others that we noted in the Shortcut story.

PR #2012, opened 2025-03-26
Cyberskills2Work Dashboard: Removed filter button when csv modal is opened

Romerico234:
There was an issue in the mobile view of the dashboard where the filter button appeared on top of the CSV generation modal and remained functional. To fix this, I used an ngIf to conditionally show or hide the filter buttons but for both desktop and mobile views. This approach removes the buttons from the background entirely, but I felt it was acceptable since with the release of this dashboard, users won’t be able to scroll/interact with the background while any modal is open.

Video:

https://github.com/user-attachments/assets/e00b9168-20b9-46ce-ae68-0dc25569b316

### DIFF
--- a/src/app/collection/cyberskills-dashboard/cyberskills-dashboard.component.html
+++ b/src/app/collection/cyberskills-dashboard/cyberskills-dashboard.component.html
@@ -5,7 +5,7 @@
 
   <div class="downloads-header">
     <h2>All Learning Objects</h2>
-    <div class="filters-wrap">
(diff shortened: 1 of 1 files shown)


pzalep1 (commented):
Can we get a video of the behavior?

Greek (approved):
this works but you could try creating a new class under the filters-wrap class, like `filters-wrap modal-open` that will add a `z-index: -999` to filters-wrap, so it will be under the modal.

johngonz0 (approved):

PR #2023, opened 2025-04-16
chore: [sc-36550] Add HTML regex filter on inputs for Learning Object Builder

Greek:
Adds a regex for non-terminating tags. Prevents users from entering `<a href="googleplex.com">asdadsasd` without a closing tag, and other html tags.

(You can't actually hit this regex in the browser (unless your browser is really old) because st-editor will automagically add the closing tag back if it doesn't find one, which is very neat.)

Check out the example: https://regex101.com/r/yWbTFg/1
Story details: https://app.shortcut.com/clarkcan/story/36550

### DIFF
--- a/src/entity/learning-object/submittable-learning-object.ts
+++ b/src/entity/learning-object/submittable-learning-object.ts
@@ -115,8 +115,8 @@ export class SubmittableLearningObject extends LearningObject {
 
 export namespace SubmittableLearningObject {
   // Matches anything that does not have a closing HTML tag.
-  // Example: https://regex101.com/r/qDSGWv/1
-  const NON_TERMINATING_HTML_REGEX = /<([a-z]+)(\s[^>]*)?(?<!\/)>(?![\s\S]*<\/\1>)/gi;
+  // Example: https://regex101.com/r/TCRXf8/1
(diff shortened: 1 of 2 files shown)


Romerico234 (approved):

src/entity/learning-object/submittable-learning-object.ts
```diff
@@ -122,12 +122,21 @@ export namespace SubmittableLearningObject {
...
+    const nonTerminatingTagRegex = /<([a-z]+)(\s[^>]*)?(?<!\/)>(?![\s\S]*<\/\1>)/gi;
+
     if (!description || !description.trim()) {
       throw new EntityError(
         SUBMITTABLE_LEARNING_OBJECT_ERRORS.INVALID_DESCRIPTION,
         'description',
       );
     }
+
+    // st-editor will always close HTML tags, so theoretically we shouldn't hit this.
```
pzalep1, line 136: This is only because the st-editor was mentioned. Would we need to catch this for the notes field too?

johngonz0 (approved):

joaquinVill (approved):

src/entity/learning-object/submittable-learning-object.ts
Greek: done

dsoto18 (approved):

PR #2042, opened 2025-05-12
Chore/sc 36139/upgrade angular from to on clark client

dsoto18:
Upgrading Clark Client Angular version from v14 to v15, most of the changed files are small changes to the scss files of different components. All the components that imported the file with all the global variables which is _vars.scss, were originally imported as @import 'vars.scss' and was allowed because of previous versions using Node Sass which had more lenient import resolution tools. Angular 15 uses Dart Sass which is stricter and cannot auto resolve vars.scss to _vars.scss. 

Onion and Learning-Object-Builder components were edited to implement AfterViewInit, they were both giving an error for Expression Changed After Checked, which was because both were using getState() functions o return activatedRouteData which was undefined at the first detection check, then changed which would throw the error. Both were fixed by using AfterViewInit and ChangeDetectorRef.

There were a couple of smaller bugs from referencing undefined values that just required a ? when using them. A couple other bugs were found that already exist in prod but do not prevent any of the necessary interactions from performing properly, some stories will be out soon eventually to fix them. Everything works fine in dev and will be tested again in staging.


src/app/onion/learning-object-builder/components/content-upload/app/upload/upload.component.html
```diff
@@ -60,7 +60,7 @@
           <st-editor
             [(ngModel)]="notes"
             [config]="config"
-            (keyup)="updateNotes($event)"
+            (keyup)="updateNotes(notes)"
```
johngonz0, line 63: is there no event for this?

johngonz0 (approved):

src/app/shared/components/input-field/input-field.component.scss
```diff
@@ -4,4 +4,12 @@
 
 mat-icon:hover{
     cursor: pointer;
-}
\ No newline at end of file
+}
+
```
dsoto18, line 8: these override the styling changes that were made in the new implementation for mat-form-fields, which was adding a box-shadow in the back of every input text box

src/app/shared/components/input-field/input-field.component.html
```diff
@@ -26,5 +25,6 @@
             authValidation.getInputErrorMessage(control, confirmType)
         }}
     </mat-error>
```
dsoto18, line 27: Updates to mat-form-field changed the way placeholders are passed through.

src/app/onion/learning-object-builder/components/content-upload/app/upload/upload.component.html
dsoto18: It looked like a bug initially in the learning object builder, how it was previously kept throwing an error when trying to update the notes of a learning object, I think the correct way was to pass through the actual string in the method call and it seemed to fix the errors from throwing

pzalep1 (changes requested):
Once the input fields are fixed this will be good to go.

dsoto18:
> Once the input fields are fixed this will be good to go.

The last two commits actually fixed it, looks the same as it did before now @pzalep1

pzalep1 (approved):

PR #2046, opened 2025-06-06
Bug Fixes to add onto the Clark Angular 15 Upgrade

dsoto18:
switched param username for userid in access groups api call, also fixed alignment of mat input form layout.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clark",
-  "version": "6.7.0",
+  "version": "6.8.0",
   "lockfileVersion": 3,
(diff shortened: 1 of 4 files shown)


dsoto18:
when testing the angular upgrade there was one bug that I saw was easily fixable, so I made the change it needed and also added some scss in the mat-input file to override styling and give the input forms the correct layout it needed after you pointed out to me it was slightly off in the login page @pzalep1 now looks good there and in the registration form

joaquinVill (approved):

pzalep1 (approved):

PR #2067, opened 2025-07-24
Bug/sc 37583/blooms taxonomy apply and analyze perform

dsoto18:
Fixes the issue Leah brought up with the blooms taxonomy 'perform' verb not saving when selected from a user creating a learning outcome in the builder. The issue was in the isGoodVerb() function and had to do with the order and method it used to traverse the package and see if the selected verb existed. Now it accounts for the selected category which is this.bloom and makes sure that matches as well. 

Uploaded a screen recording showing it saves now, the error logs in the console are all related to the downtime route still being used somewhere which is deprecated.

### DIFF
--- a/src/app/onion/learning-object-builder/components/outcome/outcome-typeahead/outcome-typeahead.component.ts
+++ b/src/app/onion/learning-object-builder/components/outcome/outcome-typeahead/outcome-typeahead.component.ts
@@ -144,7 +144,7 @@ export class OutcomeTypeaheadComponent implements OnInit, OnChanges, OnDestroy {
     if (!this.bloom) {
       const entries = Object.entries(taxonomy.taxons);
       const match = entries.find(([_, taxon]) => taxon.verbs.includes(verb.toLowerCase()));
-      
+
       if (match) {
         this.bloom = match[0];
       }


dsoto18:
https://github.com/user-attachments/assets/42a944ca-f8a0-4002-bd81-e8cd17e78c84

src/app/onion/learning-object-builder/components/outcome/outcome-typeahead/outcome-typeahead.component.ts
```diff
@@ -141,16 +141,15 @@ export class OutcomeTypeaheadComponent implements OnInit, OnChanges, OnDestroy {
...
-          this.selectedCategory.emit(this.bloom);
-        }
-        return true;
+    if (
+      this.bloom &&
+      taxonomy.taxons[this.bloom] &&
+      taxonomy.taxons[this.bloom].verbs.includes(verb.toLowerCase())
+    ) {
+      if (!noEmit) {
+        this.selectedCategory.emit(this.bloom);
```
pzalep1, line 150: If you copy an outcome into the text box can it still be recognized correctly? I noticed in the demo construct typed in was throwing an error.

dsoto18: Actually thought that was normal behavior at first, but switched to main and see that yea it shouldn't do that. I'll make some edits @pzalep1

dsoto18:
https://github.com/user-attachments/assets/3595026c-73aa-45fd-934c-7da152af754a

Updated Screen Recording

src/app/onion/learning-object-builder/components/outcome/outcome-typeahead/outcome-typeahead.component.ts
```diff
@@ -141,22 +141,33 @@ export class OutcomeTypeaheadComponent implements OnInit, OnChanges, OnDestroy {
...
-        if (!noEmit) {
-          this.selectedCategory.emit(this.bloom);
+      if (levelsArray[i] === this.bloom) {
+        if (taxonomy.taxons[levelsArray[i]].verbs.includes(verb.toLowerCase())) {
+          this.bloom = levelsArray[i];
+          if (!noEmit) {
+            this.selectedCategory.emit(this.bloom);
+          }
+          return true;
         }
```
dsoto18, line 162: all that needed to be added was an extra check to make sure this.bloom (the selected category) matches one of the levels in the taxonomy, but then another if block was needed at the start of the function to handle a bloom not being selected from the drop down in the scenario that an outcome is copy and pasted in. In that case, we only have the verb which we use to search up through the taxons structure to find the key and if it is a duplicate verb, the first category that contains it will be assigned to this.bloom

PR #2076, opened 2025-08-14
Feature/sc 37545/with cyber tags page on clark client

joaquinVill:
Adds WITHCyber tag collection page to clark client

Notes:
 - Since WITHCyber is not a collection and learning objects can only be in 1 collection, the WITHCyber page will display learning objects with the logo of the collection that they are in.
 - In the video you may notice that the stats returns 668 learning objects but when you browse, the search learning objects only returns 664. This is a backend issue that I am looking into. 
 - The RGB spec that I used in the header component is color matched to the WITHCyber red that is in the logo incase you were wondering if I randomly selected that color.
 
 Video; 
 

https://github.com/user-attachments/assets/9179b2e9-3de9-4c64-a830-b0eec6ae73fa

### DIFF
--- a/src/app/collection/pages/tags/tag-with-cyber/tag-with-cyber.component.ts
+++ b/src/app/collection/pages/tags/tag-with-cyber/tag-with-cyber.component.ts
@@ -1,11 +1,11 @@
 import { Component, OnDestroy, OnInit } from '@angular/core';
 import { Collection, LearningObject } from '@entity';
 import { CollectionService } from 'app/core/collection-module/collections.service';
-import { FeaturedObjectsService } from 'app/core/feature-module/featured.service';
 import { NavbarService } from 'app/core/client-module/navbar.service';
-import { Title } from '@angular/platform-browser';
-import { SEARCH_ROUTES } from 'app/core/learning-object-module/search/search.routes';
 import { TagsService } from 'app/core/learning-object-module/tags/tags.service';
+import { SearchService } from 'app/core/learning-object-module/search/search.service';
+import { OrderBy } from 'app/interfaces/query';
+
 
 @Component({
   selector: 'clark-tag-with-cyber',
@@ -21,9 +21,8 @@ export class TagWithCyberComponent implements OnInit, OnDestroy {
   constructor(
     private navbarService: NavbarService,
     private collectionService: CollectionService,
-    private titleService: Title,
-    private featureService: FeaturedObjectsService,
-    private tagsService: TagsService
+    private tagsService: TagsService,
+    private searchService: SearchService
   ) {}
 
   async ngOnInit() {
@@ -32,8 +31,6 @@ export class TagWithCyberComponent implements OnInit, OnDestroy {
     this.collection = await this.collectionService.getCollection(this.abvCollection);
 
     this.learningObjects = await this.getFeaturedLearningObjects();
-
-    this.titleService.setTitle('CLARK | ' + this.collection.name);
 
     this.tagId = await this.tagsService.getTagIdByName('WITHcyber');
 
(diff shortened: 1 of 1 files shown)


src/app/collection/pages/collection-ncyte/components/stats/stats.component.ts
```diff
@@ -131,8 +145,30 @@ export class StatsComponent implements OnInit {
...
+
+  async getWITHCyberTagId() {
+      const url =  TAGS_ROUTES.GET_ALL_TAGS({ text: 'WITHCyber' });
+      const res = await fetch(url, { method: 'GET' });
+      const data: TagsResponse = await res.json();
+      const tag =  data.tags?.[0]?._id ?? null;
+      console.log('Tag ID: ', tag);
+      return tag;
+    }
+}
```
dsoto18, line 171: I think we can clean up this function a bit, instead of importing the url from TAGS_ROUTES and performing the fetch request like we have here, it'd probably be best to use the getTags() function in the tags.service.ts file. That tags service has all the fetch request code abstracted in there so we can reuse it, see tags.component.ts line 55 as an example, it does the same thing passing a single text query.

src/app/collection/pages/tags/tag-with-cyber/components/header/header.component.ts
```diff
@@ -0,0 +1,38 @@
+import { Component, Input } from '@angular/core';
+import { Router } from '@angular/router';
+import { TAGS_ROUTES } from 'app/core/learning-object-module/tags/tags.routes';
+
+interface TagsResponse {
+  tags: { _id: string }[];
+  total: number;
+}
```
dsoto18, line 8: I saw this same interface was made in the stats component, we can probably throw it in the usage-stats.ts file that has the other interfaces and reference it from there.

src/app/core/metric-module/metric.service.ts
```diff
@@ -74,6 +75,14 @@ export class MetricService {
...
   }
 
+  async getTagMetrics(tagId?: string): Promise<TagMetrics> {
+    const tagMetrics: TagMetrics = await this.http
+      .get<TagMetrics>(METRIC_ROUTES.GET_TAG_METRICS(tagId))
+      .pipe(catchError(this.handleError))
+      .toPromise();
+    return tagMetrics as TagMetrics;
+  }
+
```
dsoto18, line 85: Exactly like you did this here is what we can do for the searchLearningObjects and the getTags requests I left comments on above! Those services already have the implementation but this is the same idea of having these requests in the service files, then calling that in the component files

dsoto18:
Really nice work here, just wondering though - can we double check that the other collection pages that use the same stats component still look normal? Just want to make sure since I see we're using the same component/file for this tags page. I think the story kinda mentioned to create a new stats component similar to the ncyte one, but if you made it work using the same one then its all good! Just think its a good idea to double check the other ones still appear as they did before.

src/app/core/learning-object-module/tags/tags.service.ts
```diff
@@ -117,6 +118,15 @@ export class TagsService {
       .toPromise();
   }
 
+  public async getTagIdByName(tagName: string): Promise<string> {
+     const url =  TAGS_ROUTES.GET_ALL_TAGS({ text: tagName });
+        const res = await fetch(url, { method: 'GET' });
+        const data: GetTagByNameResponse = await res.json();
+        const tag =  data.tags?.[0]?._id ?? null;
+        console.log('Tag ID: ', tag);
```
curlyLasagna, line 126: 53bc1a98-29ce-4d40-be37-44218aa62a6e

curlyLasagna (approved):

src/app/collection/pages/tags/tag-with-cyber/components/header/header.component.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Router } from '@angular/router';
+import { TagsService } from 'app/core/learning-object-module/tags/tags.service';
+@Component({
+  selector: 'clark-with-cyber-header',
+  templateUrl: './header.component.html',
+  styleUrls: ['./header.component.scss']
+})
+export class HeaderWithCyberComponent {
+
+  @Input() collectionAbv: string;
```
dsoto18, line 11: nit, since this component is for the header of just the with-cyber component, can we name this input value to be tagName, or something instead of collectionAbv since that's not really what it will be?

src/app/collection/pages/tags/tag-with-cyber/tag-with-cyber.component.ts
```diff
@@ -0,0 +1,61 @@
...
+      return [];
+    }
+    // for now we are just fetching the most recent 5 LOs for the featured section of the collection page
+    const queryParams = new URLSearchParams({ tags: tagId, orderBy: 'date', sortType: '-1',  limit: '5' }).toString();
+    const url = SEARCH_ROUTES.SEARCH_LEARNING_OBJECTS(queryParams);
+    const res = await fetch(url, { method: 'GET' });
+    const payload: { objects?: LearningObject[]; learningObjects?: LearningObject[]; total?: number } = await res.json();
+    const list = payload.objects ?? payload.learningObjects ?? [];
+    const top5 = list.slice(0, 5);
+    return top5;
```
dsoto18, line 59: there's no search service file with a getLearningObjects function to use instead of this?

PR #2087, opened 2025-09-16
Feature/sc 37676/use standard guidelines search guidelines

curlyLasagna:
- Adds the search guidelines component from the builder to the tagging builder. The result of the search can be clicked to map the expanded outcome
- Adds a suggested guidelines column for alignment. The suggestion is keyword based. Keywords being the concatenation of a learning object's name, description and its list of learning outcomes.
- Adds the ability to remove a mapping from a learning outcome
- Created a stripHTML shared function
- Modified the styling of the tagging builder popup to be wider to accommodate the 3 components within the alignment tab

https://github.com/user-attachments/assets/e1e2bec1-92d0-4454-9774-516fef32270e


src/app/cube/details/components/action-panel/editorial-action-pad/components/guidelines/guidelines.component.scss
```diff
@@ -1,25 +1,61 @@
...
+}
+
+/*
+https://stackoverflow.com/questions/47024236/what-to-use-in-place-of-ng-deep
+I like Simon_Weaver's answer to the question of whether to use ng-deep or not
+
+angular chips' text are set to not wrap and I want it to wrap
+*/
+:host ::ng-deep .mdc-evolution-chip__text-label {
+  white-space: normal;
```
curlyLasagna, line 60: https://github.com/Cyber4All/clark-client/pull/2080

Similar to the above PR `:host ::ng-deep` can be used to overwrite 3rd party component styling

joaquinVill (approved):

dsoto18:
Everything looks good to me, nice work! Just a couple questions - 

I think we talked about it already and I saw it in the code.. but so if there are no suggested guidelines to display in that middle column it will show the 8 core KSATs right? Will it display a message about that (like something that says "no results, here are the 8 core ksats")? Also what if there are only 1 or 2 suggested guidelines, will only those 2 be returned and nothing to fill in the gap for the rest of the white space? Not an issue if it does do that, just curious.

Also, when you click a guideline to map it to a learning outcome, a learning outcome (from the left column) must be open, right? Like that's how it knows which outcome to map to, whichever outcome that is opened with the dropdown is the 'active' one? 

Also I don't believe this should be happening but just to be safe, is the current search by guidelines in the clark builder still working as normal with these changes? @curlyLasagna

curlyLasagna:
@dsoto18 
> Also, when you click a guideline to map it to a learning outcome, a learning outcome (from the left column) must be open, right? Like that's how it knows which outcome to map to, whichever outcome that is opened with the dropdown is the 'active' one?

Yep, it shouldn't change the previous that it had before when it comes to mapping an outcome

> Also I don't believe this should be happening but just to be safe, is the current search by guidelines in the clark builder still working as normal with these changes?

Tested it out and works as before but we can do further testing in staging to confirm

> but so if there are no suggested guidelines to display in that middle column it will show the 8 core KSATs right?

Good catch, I forgot an `else`
```ts
if (result.results.length === 0) {
      // Use the hard coded 8 core KSATs when there are no results
      this.alignmentService.setGuidelinesArray(selectedGuidelines as SearchItemDocument[]);
    }
    this.alignmentService.setGuidelinesArray(result.results);
```

dsoto18 (approved):

PR #2096, opened 2025-10-03
Disable sortBy date as default value

curlyLasagna:
Sort by date is set as the default value on browse component creation, so I set `orderBy` to an empty string to avoid sorting by date by default. 

I'm going to assume that the intent is to show the newest learning objects when a user visits the browse page.

Should we change this behavior?

After performing a search query, the `sortBy` and `sortType` query param is still being passed, so the exact match is never at the top of the search results since sorting by date gets prioritized. 

It's evident when you search then go to the next page as the URL gets updated to show those query params.

### DIFF
--- a/src/app/cube/browse/browse.component.ts
+++ b/src/app/cube/browse/browse.component.ts
@@ -34,7 +34,7 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
     guidelines: [],
     level: [],
     standardOutcomes: [],
-    orderBy: OrderBy.Date,
+    orderBy: '',
     sortType: -1,
     collection: '',
     topics: [],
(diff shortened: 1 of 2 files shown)


Cwagne17 (approved):

src/app/components/search/search.component.ts
```diff
@@ -108,6 +108,8 @@ export class SearchComponent implements OnInit, OnDestroy {
           ...this.route.snapshot.queryParams,
           text,
           currPage: 1,
+          // Disable any other sorting
+          orderBy: ''
```
dsoto18, line 112: sorry for not understanding this as clearly, but what would be the purpose of this line here?

src/app/cube/browse/browse.component.ts
```diff
@@ -86,7 +87,6 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
     private router: Router,
     private cd: ChangeDetectorRef,
     private navService: NavbarService,
-    private searchService: SearchService,
```
dsoto18, line 89: also just wondering, was search service not called at all in this file?

PR #2176, opened 2026-02-20
bug/sc-38722/lo-topics-latency-bug

dsoto18:
# Description
Refactored the implementation of retrieving topics and tags for a learning object to utilize a caching strategy. Introduces two new files `topics.store.ts` and `tags.store.ts` which service as the caching layer, which is separate from the actual network layer in their respective service files.

One thing to note is there is no single GET /topicById route, so for Topics the strategy workflow is when a topic is needed -> perform GET all tags request and store them in static map -> then in the future when a topic is needed get that value from the map and not perform another API request -> and in the unlikely (realistically impossible) event that a topic does not exist in the map -> perform the GET all tags request again. 

Also we've had a GET single tag by tagId route in our API, but it was never used in our client until now. I introduced it here so we can have an actual cache with the same functionality as the organizations, even though there's only ~30 tags, but still eliminates grabbing tags we don't need which is the purpose of a cache. I added it to the tag routes file for now, but I know if the future we may have a story to migrate away from this pattern. 

This will reduce the amount of network requests improving the latency issue we were seeing on the browse component, and also help us clean up our logs a bit. 

Screen recording available here: https://securityinjections.slack.com/archives/C02PFHGANF5/p1771615411862139


src/app/cube/shared/featured/featured.component.ts
```diff
@@ -57,8 +57,12 @@ export class FeaturedComponent implements OnInit {
         this.query
       )).learningObjects;
       this.loading = false;
+      this.cd.detectChanges();
     } catch (e) {
+      console.log("caught error");
```
curlyLasagna: 🔥

src/app/cube/shared/featured/featured.component.ts
```diff
@@ -26,27 +28,21 @@ export class FeaturedComponent implements OnInit {
...
-          this.query.collection = collection;
-          this.fetchLearningObjects();
-        }
-      });
-    } else {
-      this.featureService.getFeaturedObjects();
-      this.featureService.featuredObjects.subscribe(objects => {
-        this.learningObjects = objects;
-      });
-    }
```
dsoto18: the featured coll in mongo used to hold additional learning objects that weren't meant to be featured on the home page, but could be called in here as a 'backup' (i'm assuming), but I know there was no way to add collection-specific featured LOs through featured service, and it used to be done manually by Mike and Paige. Either way this else block is never getting hit,

curlyLasagna (approved):

joaquinVill (commented):
looks good to me nice job

Cwagne17 (commented):
Looks good. I'd recommend checking out the store and service split to slightly adjust the caching algorithm.

Also does the browse filters page need to adjust how it uses topics/tags?

Opt: Don't have to necessarily do the strictly typed schemas right now. Might want to at least put TODOs to use zod in the future to type the schemas for response, request, and types.

Cwagne17 (approved):

PR #2181, opened 2026-03-02
Replaces the back button to say "Save & Return"

curlyLasagna:
Replaces the back button to say "Save & Return" to let the users know that their work is saved.

image


dsoto18 (approved):
Nice, just wondering - were you able to test this button locally? I think I remember recently when I was deving here it would give an error on the back button, cause it would try to hit our S3 driver in dev mode.

joaquinVill (approved):

PR #2209, opened 2026-04-24
Add organization migration capability to admin page

Cwagne17:
This PR adds the Organization Migration capability to the Admin Organization page. This enables an Admin to migrate an Organization's user's and learning objects to another Verified organization in just a few clicks. The search for organizations filters for only verified organizations and requires an acknowledgement before migrating.

**THIS IS IRREVERSIBLE** - reverting a migration would be difficult to track which objects and users were migrated without reverting to a previous DB snapshot.

https://github.com/user-attachments/assets/e5aed7f8-cd2d-4d60-9a04-ce750a456f64


src/app/admin/pages/organizations/organizations.component.spec.ts
```diff
@@ -1,99 +1,128 @@
...
+import { of, throwError } from 'rxjs';
 import { OrganizationsComponent } from './organizations.component';
 
 describe('OrganizationsComponent', () => {
   let component: OrganizationsComponent;
-  let toasterService: ToastrOvenService;
-  let organizationService: OrganizationService;
-  let userService: UserService;
-  let searchService: SearchService;
+  let toaster: ReturnType<typeof jasmine.createSpyObj<ToastrOvenService>>;
```
dsoto18, line 13: do we have this jasmine createSpy implementation somewhere already in clark, just curious

dsoto18 (approved):
lgtm, I like the caching strategy to figure out learning object counts per organization. How much attention should we pay to the sonar cloud warnings too by the way?

joaquinVill (approved):

Cwagne17:
Up to you, we don't have to use them, I was just checking it out as a potential security solution for our code.

PR #2210, opened 2026-04-24
chore/sc-39006/update-organization-not-listed-form

dsoto18:
## Summary

This PR improves the registration flow for users who need to manually enter their organization information.

### What changed

- Replaced free-text `Country` and `State` inputs on the organization step with dropdowns
- Added a feature-local constants file with country options and U.S. state options
- Made `Country` required
- Made `State` required only when `United States` is selected
- Kept `Education Levels` optional
- Swapped the field order so `Country` appears on the left and `State` on the right
- Updated organization-step validation so clicking `Next` surfaces all missing required-field errors, even if the user has not interacted with those dropdowns yet
- Added a tooltip to the disabled “Organization not listed, enter my organization info” link using the app’s shared `[tip]` directive, with the message:
  `Enter first name, last name, and email first`

### Notes

- Country values are stored as country codes internally while showing full country names in the UI
- The `State` dropdown is visible at all times but disabled unless `United States` is selected
- Country and state values are hardcoded since these values will likely not be changing often

https://github.com/user-attachments/assets/0b1b1ec2-8abf-477a-b985-812861ae7309


src/app/auth/register/register-location-options.ts
```diff
@@ -0,0 +1,304 @@
+export interface SelectOption {
+  value: string;
+  label: string;
+}
+
+export const COUNTRY_OPTIONS: SelectOption[] = [
+  { value: 'US', label: 'United States' },
```
dsoto18, line 7: Having the US first here does make it appear as the first of the drop down options

joaquinVill (approved):

PR #2215, opened 2026-05-19
chore/sc-39281/bring-back-sort-by-oldest-on-clark-for-admins

dsoto18:
# Summary
Asked by Leah in the dev-ops channel, but adding back sort by Oldest first in the browse page as an admin only feature. Before this was removed along with sort by alphabetical since those are rarely if ever used by end users. The use case for bringing back Oldest First is it would make it easier for Leah to look at what objects are going to need relevancy checks the most, and she asked to keep this admin only.

**Note:** only requires client side changes to pass the query through on the request, service already is setup to handle sort by oldest. It doesn't have admin only verification on it though, but I don't think it's a big deal really. So just these changes are needed

https://github.com/user-attachments/assets/c644bfa9-6f49-40a4-951d-3051a7e8e47d

### DIFF
--- a/src/app/cube/browse/browse.component.ts
+++ b/src/app/cube/browse/browse.component.ts
@@ -18,6 +18,7 @@ import { OrderBy, Query, SortType } from '../../interfaces/query';
 import { COPY } from './browse.copy';
 import { FilterSectionInfo } from './components/filter-section/filter-section.component';
 import { FilterComponent } from './components/filter/filter.component';
+import { AuthService } from 'app/core/auth-module/auth.service';
 
 @Component({
   selector: 'cube-browse',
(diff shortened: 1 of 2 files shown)


src/app/cube/browse/browse.component.ts
```diff
@@ -450,6 +452,10 @@ export class BrowseComponent implements AfterViewInit, OnDestroy {
         this.sortText = 'Newest';
         this.query.orderBy = OrderBy.Date;
         this.query.sortType = SortType.Descending;
+      } else if (val === 'ad') {
```
dsoto18, line 455: I think 'Newest' has the value 'dd' for 'descending date', so I gave oldest 'ad' for 'ascending date'

joaquinVill (approved):

curlyLasagna (approved):

PR #2224, opened 2026-06-09
feature/sc-38831/remove-legacy-organization-field-from-clark

dsoto18:
# Summary
This is the client PR for the organizations epic cleanup, getting rid of the now legacy implementation of holding the 'organization' string value directly on a user document. 

The only location where it still remained in the client code base was in the user profile's edit modal, organization was still being sent along with organizationId to keep things backwards compatible. It's now safe to remove and the code has been tested with the required clark-service changes, all works as it should. Those service changes are located at https://github.com/Cyber4All/clark-service/pull/507 and remove all the old logic from the api which is a lot more than here. The service changes handles the logic with changing authors on learning objects, searching learning objects, and a few other things. Client changes were a lot more simple.

The legacy organization field also used to be used in the client's registration flow but it was already removed in #2204


Low-Key-N (approved):
Looks good!

curlyLasagna (approved):

Cwagne17 (approved):

PR #2240, opened 2026-07-19
Bug/sc 38715/error adding learning object to library popup

imrakeshgulla:
## Description

This pull request addresses an issue where duplicate "Save to Library" errors or unexpected response statuses (such as a `409 Conflict`) would trigger error toast notifications or block users from seeing a correct saved state.

It introduces helper methods to properly parse duplicate error signatures, leverages a `savedLearningObjectKeys` `Set` to track saved items via their `cuid` and `version`, and updates the action panel UI to handle pending states gracefully and suppress redundant error popups.

### Library Service (`library.service.ts`)

* **Added `AddToLibraryResult` Interface:** Return typed status and an `alreadySaved` boolean indicator rather than raw data.
* **Unique Tracking:** Introduced `savedLearningObjectKeys` (`Set<string>`) to cache unique composite keys (`cuid:version`) for objects in the library.
* **Error Catching & Parsing:**
* Intercepts errors and maps `409` conflict states to a successful mock response or handles text/body checks via a new `isAlreadySavedError()` helper.
* The `isAlreadySavedError()` method checks the error message and body strings for keywords like `"already"`, `"library"`, `"saved"`, and `"duplicate"`.

* **State Alignment:** Updates the unique keys set during standard retrieval (`getLibrary`) and deletion (`removeFromLibrary`).

### Action Panel Component (`action-panel.component.ts`)

* **Concurrency Guard:** Added a private `addToLibraryPending` flag to prevent duplicate simultaneous clicks/triggers.
* **Optimized Logic Flows:** Streamlined `addToLibrary()` to explicitly evaluate `shouldAddToLibrary` using new visibility and release criteria rules.
* **Toast Suppression:** Wrapped the success toaster so it only alerts the user if the item wasn't *already* flagged as saved (`!result?.alreadySaved`), eliminating redundant toast messages.
* **Safe Status Refreshing:** Added a private `tryRefreshLibraryStatus()` helper wrapped in a try/catch to ensure a failed status refresh doesn't falsely signal an add-to-library failure.

---

## Related Commits

* `6b26e77` - Fixed the error savetolibrary
* `56a9ab5` - Suppress duplicate library save error toast

### DIFF
--- a/src/app/core/library-module/library.service.ts
+++ b/src/app/core/library-module/library.service.ts
@@ -150,7 +150,7 @@ export class LibraryService {
             )
             .pipe(
                 catchError((error) => {
-                    if (error.status === 409) {
+                    if (this.isAlreadySavedError(error)) {
                         return of(new HttpResponse({ status: 409 }));
                     }
                     return throwError(error);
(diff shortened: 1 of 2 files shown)


Low-Key-N (approved):

curlyLasagna:
Hold off from merging for now. I want to compare this client fix to a backend fix

Cwagne17 (approved):

